PR #35, opened 2020-09-08
Project CleanUp with Clippy

shekohex:
I made a Cleanup in a lot of places, using Clippy, and apply any suggestions and lints.

### DIFF
--- a/crates/rune-cli/src/main.rs
+++ b/crates/rune-cli/src/main.rs
@@ -191,41 +191,46 @@ async fn main() -> Result<()> {
     let mut sources = rune::Sources::new();
     let mut warnings = rune::Warnings::new();
 
-    let unit = loop {
-        if options.bytecode && should_cache_be_used(&path, &bytecode_path)? {
-            log::trace!("using cache: {}", bytecode_path.display());
-            let f = fs::File::open(&bytecode_path)?;
-
-            match bincode::deserialize_from::<_, Unit>(f) {
-                Ok(unit) => break Arc::new(unit),
-                Err(error) => {
-                    log::error!(
-                        "failed to deserialize: {}: {}",
-                        bytecode_path.display(),
-                        error
-                    );
-                }
+    let use_cache = options.bytecode && should_cache_be_used(&path, &bytecode_path)?;
+    let maybe_unit = if use_cache {
+        let f = fs::File::open(&bytecode_path)?;
+        match bincode::deserialize_from::<_, Unit>(f) {
+            Ok(unit) => {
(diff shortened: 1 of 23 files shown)


crates/runestick/src/generator.rs
```diff
@@ -20,14 +33,6 @@ impl Generator {
         }
     }
 
-    /// Get the next value produced by this stream.
-    pub fn next(&mut self) -> Result<Option<Value>, VmError> {
```
udoprog, line 24: This would be used in modules to iterate over a generator, and they usually have a signature which includes `Result<T, VmError>`. Moving this impl into `Iterator` would just result in every call having to "untranspose" for no benefit.

I'd suggest adding an ignore for the lint, or renaming the function. But `next` is good name, so I'm leaning on ignoring the lint.

crates/rune/src/ast/item.rs
```diff
@@ -10,6 +10,9 @@ pub enum Item {
     /// A use declaration.
     ItemUse(ast::ItemUse),
     /// A function declaration.
+    // large size difference between variants
+    // we should box this variant.
+    // https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
```
udoprog, line 15: Feel free to box the variant!

udoprog (changes requested):
Thanks, I like getting these done once in a while!

crates/rune-cli/src/main.rs
```diff
@@ -191,41 +191,46 @@ async fn main() -> Result<()> {
     let mut sources = rune::Sources::new();
     let mut warnings = rune::Warnings::new();
```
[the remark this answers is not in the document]
shekohex: Honestly, the main function is very big and a bit missy, and it should be separated into smaller functions

also, you are parsing cli options manually, if you are concerned about compile-time and bin size, we could use [`argh`](https://lib.rs/crates/argh) I'm using it in all of my cli tools.

crates/rune/src/ast/item.rs
shekohex: That would require a lot of changes in a lot of places .. so I guess we could let that for another cleanup.

crates/rune-cli/src/main.rs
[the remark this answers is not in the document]
udoprog: Sounds good to me. Do the clippy lints here, and if you want to improve the cli with better arg parsing after I'd be all for it!

udoprog:
@shekohex Merged in current state to make sure you're not blocked. Feel free to do the improvements you proposed!

PR #36, opened 2020-09-08
[CI] Cache Cargo build

shekohex:
By using Cache it would make it faster for testing and building.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,10 +8,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
+        id: rust_toolchain
         with:
           toolchain: stable
           override: true
-      - run: echo "::set-env name=RUSTC_VERSION::${{ rustc --version }}"
       - name: Cache Cargo
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -6,9 +6,18 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Cache Cargo
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ~/.cargo/bin
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
udoprog, line 18: Hm, not so familiar with this, but would the following be useful to include in the key:
* Compiler version, otherwise the cache could grow indefinitely in size between releases.
* `toolchain`, since stable won't benefit from a cached nightly bulid and vice-versa.

Cargo lock is also not checked in. So not sure it makes a difference here.

shekohex: Will update that shortly.

shekohex: Fixed, we are using `rustc` hash now
see https://github.com/actions-rs/toolchain#outputs

udoprog:
Ok, let's give it a go!

Thanks for the change, TIL about `steps.rust_toolchain.outputs` 😄!

PR #51, opened 2020-09-10
improve Rune CLI.

shekohex:
I made some improvements to the Rune CLI by using [`structopt`](https://lib.rs/crates/structopt) for parsing cli args, also some other clippy lints fixed on the way :).

### DIFF
--- a/crates/rune-cli/Cargo.toml
+++ b/crates/rune-cli/Cargo.toml
@@ -21,6 +21,7 @@ codespan-reporting = "0.9.5"
 anyhow = "1.0.32"
 bincode = "1.3.1"
 log = "0.4.11"
+structopt = "0.3"
 
 rune = {version = "0.6.16", path = "../rune", features = ["modules"]}
 rune-macros = {version = "0.6.16", path = "../rune-macros"}
--- a/crates/rune-cli/src/main.rs
+++ b/crates/rune-cli/src/main.rs
@@ -57,7 +57,7 @@ use structopt::StructOpt;
 use runestick::{Item, Unit, Value, VmExecution};
 
 #[derive(Default, Debug, Clone, StructOpt)]
-#[structopt(name = "rune", about = "The Rune Programming Language")]
+#[structopt(name = "rune", about = "The Rune Language")]
 struct Args {
     /// Provide detailed tracing for each instruction executed.
     #[structopt(short, long)]
@@ -72,6 +72,7 @@ struct Args {
     #[structopt(long)]
     dump_instructions: bool,
     /// Dump the state of the stack after completion.
+    ///
     /// If compiled with `--trace` will dump it after each instruction.
     #[structopt(long)]
     dump_stack: bool,
@@ -90,28 +91,26 @@ struct Args {
     /// Include source code references where appropriate (only available if -O debug-info=true).
     #[structopt(long)]
     with_source: bool,
-    /// Enabled experimental features.
+    /// Enable experimental features.
+    ///
+    /// This makes the `std::experimental` module available to scripts.
     #[structopt(long)]
     experimental: bool,
     /// Input Rune Scripts
     #[structopt(parse(from_os_str))]
     paths: Vec<PathBuf>,
-    ///  Update the given compiler option.
-    /// memoize-instance-fn[=<true/false>]
-    ///     Inline the lookup of an instance function where appropriate.
+    /// Set the given compiler option (see `--help` for available options).
     ///
-    /// link-checks[=<true/false>]
-    ///     Perform linker checks which makes sure that called functions exist.
+    /// memoize-instance-fn[=<true/false>] - Inline the lookup of an instance function where appropriate.
     ///
-    /// debug-info[=<true/false>]
-    ///     Enable or disable debug info.
+    /// link-checks[=<true/false>] - Perform linker checks which makes sure that called functions exist.
     ///
-    /// macros[=<true/false>]
-    ///     Enable or disable macros (experimental).
+    /// debug-info[=<true/false>] - Enable or disable debug info.
     ///
-    /// bytecode[=<true/false>]
-    ///     Enable or disable bytecode caching (experimental).
-    #[structopt(short = "O")]
+    /// macros[=<true/false>] - Enable or disable macros (experimental).
+    ///
+    /// bytecode[=<true/false>] - Enable or disable bytecode caching (experimental).
+    #[structopt(name = "option", short = "O")]
     compiler_options: Vec<String>,
 }
 
--- a/crates/rune-cli/src/main.rs
+++ b/crates/rune-cli/src/main.rs
@@ -48,144 +48,107 @@
 use anyhow::Result;
 use rune::termcolor::{ColorChoice, StandardStream};
 use rune::EmitDiagnostics as _;
-use std::env;
 use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use structopt::StructOpt;
 
 use runestick::{Item, Unit, Value, VmExecution};
 
-#[derive(Default, Debug, Clone)]
+#[derive(Default, Debug, Clone, StructOpt)]
+#[structopt(name = "rune", about = "The Rune Programming Language")]
 struct Args {
-    path: PathBuf,
+    /// Provide detailed tracing for each instruction executed.
+    #[structopt(short, long)]
     trace: bool,
+    /// Dump everything.
+    #[structopt(short, long)]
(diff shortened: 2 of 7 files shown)


crates/rune-cli/Cargo.toml
```diff
@@ -21,7 +21,7 @@ codespan-reporting = "0.9.5"
 anyhow = "1.0.32"
 bincode = "1.3.1"
 log = "0.4.11"
-
+argh = "0.1.3"
```
udoprog, line 24: I keep the internal deps one line separate, because it makes it easier to verify version bumps during releases. Please keep it for now.

crates/rune/src/options.rs
```diff
@@ -51,25 +51,35 @@ impl Options {
     }
 
     /// Set if debug info is enabled or not. Defaults to `true`.
-    pub fn debug_info(&mut self, enabled: bool) {
+    pub fn debug_info(&mut self, enabled: bool) -> &mut Self {
```
udoprog, line 54: Method chaining can be a problematic approach to using an API, see:

https://randompoison.github.io/posts/returning-self/

It hits a few notes that will be relevant here, b so I'd rather you avoid it. And there's no reason to `cascade!` yet either. It's not complex enough.

crates/rune-cli/src/main.rs
```diff
@@ -45,159 +45,192 @@
...
+    #[argh(switch)]
+    dump_native_types: bool,
+    /// include source code references where appropriate (only available if -O debug-info=true).
+    #[argh(switch)]
+    with_source: bool,
+    /// enabled experimental features.
+    #[argh(switch)]
+    experimental: bool,
+    #[argh(subcommand)]
+    compiler_options: CompilerOptions,
```
udoprog, line 100: It's like you to figure out a way to stick with using `Options::parse_option`, or similar pattern. The compiler is intended to be integrated at a low level into other projects, and I'd rather only maintain options parsing in one location in case they want to provide a pass through mode.

I realize that this gives better diagnostics (these are the options available). And I'd like improvements here as well.

udoprog (changes requested):

udoprog:
Looks good, with a few opinions on how Options should be used.

crates/rune-cli/src/main.rs
shekohex: IMO, Keeping `CompilerOptions` as a subcommand offers a lot of advantages:
1. It is easier to pass options without typo (the CLI Parser will refuse it directly)
2. it is easy to add new options .. just one line to the subcommand and the parser handle the rest.
3. It offers a better and organized documentation and a help message

udoprog: Yes, but those benefit can't be shared by other subsystems that will need to parse compile options, these include (but are not limited to):
* Compiler options passed to the language server.
* A future package manager.
* Third party projects embedding the compiler.

So everything we do related to parsing compiler options has to be integrated in these environments, and keeping two things in sync is error prone, so I want to avoid that. Adding the help text in the current cli was already hacky 😅.

I'd suggest you add a generic help message and a separate command which lists compiler options that uses something from `Options` to get help about the list of available options.

shekohex: Well, if we implemented `FromStr` to the `Options` we could use it directly as subcommand without any wrappers. I will do this update now.

udoprog (changes requested):
Just to clarify two things missing at this point:
* The ability to specify multiple `-O` options (unless I'm missing something). Not strictly necessary, but this is what I'm used to!
* The ability to list the available options (which was previously shown in `--help`).

shekohex:
> The ability to specify multiple -O options (unless I'm missing something). Not strictly necessary, but this is what I'm used to!

You could specify multiple -O options by separate them by `,` for example

```
$ rune -O debug-info=true,macros=false
```

> The ability to list the available options (which was previously shown in --help).

the current help list shows the for the `-O` all options as comments.

```
rune --help
```
```
Usage: rune.exe <path> [--trace] [-d] [--dump-unit] [--dump-instructions] [--dump-stack] [--dump-functions] [--dump-types] [--dump-native-functions] [--dump-native-types] [--with-source] [--experimental] -O <compiler-options>

Rune Programming Language. CLI Arguments

Options:
  --trace           provide detailed tracing for each instruction executed.
  -d, --dump        dump everything.
  --dump-unit       dump default information about unit.
  --dump-instructions
                    dump unit instructions.
  --dump-stack      dump the state of the stack after completion. If compiled
                    with `--trace` will dump it after each instruction.
  --dump-functions  dump dynamic functions.
  --dump-types      dump dynamic types.
  --dump-native-functions
                    dump native functions.
  --dump-native-types
                    dump native types.
  --with-source     include source code references where appropriate (only
                    available if -O debug-info=true).
  --experimental    enabled experimental features.
  -O, --compiler-options
                    update the given compiler option (seprated by ",").
                    link-checks: Perform link-time checks, memoize_instance_fn:
                    Memoize the instance function in a loop, debug_info: Include
                    debug information when compiling, macros: Support
                    (experimental) macros, bytecode: Support (experimental)
                    bytecode caching,
  --help            display usage information


```

udoprog:
> You could specify multiple -O options by separate them by , for example

Yeah, so please make it so you can specify multiple options. I'm not familiar with `argh`, but *I'm guessing* it can be done by using `Vec<String>` as an argument type? If so, it could be straight up fed to the existing `parse_option` impl.

> the current help list shows the for the -O all options as comments.

Ok, so it picks up the doc comment! That works for me.

crates/rune/src/options.rs
```diff
@@ -14,6 +17,19 @@ pub struct Options {
...
+impl FromStr for Options {
+    type Err = ConfigurationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut opt = Options::default();
+        let it = s.split(',');
+        for o in it {
+            opt.parse_option(o)?;
+        }
+        Ok(opt)
```
shekohex, line 29: @udoprog this the code related to parsing the options separated by `,` so there is no need to make it as `Vec`.

shekohex:
@udoprog Can we get that merged before it gets more Conflicting files?

udoprog:
> @udoprog Can we get that merged before it gets more Conflicting files?

Sure, I want you to add support for specifying multiple `-O` options to the CLI. The branch (as it stood during the review) failed with this:

```text
$> cargo run -- scripts/test.rn -O debug-info=true -O debug-info=false
Error parsing option '-O' with value 'debug-info=false': duplicate values provided
```

I am aware that it's possible to specify multiple options that are comma-separated with you change, but that is a change in behavior I don't want to enforce right now. At some point compiler options can get complex enough that they'll need their own comma-separated lists which could conflict with the parsing you proposed.

I also noticed that `-O` is now a required option. Which it shouldn't be:

```text
Required options not provided:
    --compiler-options

error: process didn't exit successfully: `target\debug\rune.exe scripts/test.rn` (exit code: 1)
```

To be clear, if this change only introduced `argh` without changing any behavior I wouldn't mind merging it. It's unfortunate that it got tangled up with other clippy changes, since they seem to be causing merge conflicts. Keeping them separate in the future would be a good thing!

udoprog:
Will echo here the issue found on Discord:

The one thing that doesn't seem to work at all with `argh` is multiline documentation. The documentation string for compiler options is unreadable, like this:

```text
  -O, --compiler-options
                    update the given compiler option. link-checks: Perform
                    link-time checks, memoize_instance_fn: Memoize the instance
                    function in a loop, debug_info: Include debug information
                    when compiling, macros: Support (experimental) macros,
                    bytecode: Support (experimental) bytecode caching,
```

The only thing I added to the branch was more documentation that seemed to be missing. But it makes the squashing worse. argh might not be a good option for us if this can't be fixed, but please have a try!

crates/rune-cli/src/main.rs
```diff
@@ -48,144 +48,103 @@
...
+            || args.dump_native_functions
+            || args.dump_stack
+            || args.dump_types
+            || args.dump_instructions
+        {
+            args.dump_unit = true;
+        }
+        args
+    };
+    let options = args.compiler_options.join(",").parse()?;
```
udoprog, line 145: Iterate over the options and parse them. Otherwise we're parsing commas again (see [this comment](https://github.com/rune-rs/rune/pull/51#issuecomment-691481403)).

This also means the [FromStr impl](https://github.com/rune-rs/rune/pull/51/files#diff-aca70ad2ef68c58ca9535c2da12b8ab3R20) is not needed.

udoprog (approved):

PR #53, opened 2020-09-11
Reduced loop{} overhead

genusistimelord:
I believe this will reduce the overhead of loop{} a little bit. Did not seems to need to do a Unit return either as the Tests functioned correctly.


udoprog:
Makes sense, loop is infinite until `break` is called. And cleanup is performed according to criteria specified in `needs` at the `break`.

Good stuff!

PR #54, opened 2020-09-12
HTTP module: Use rustls, enable gzip

killercup:
This eliminates the dependency on OpenSSL and along the way also
enables support for deflate decompression of requests.

### DIFF
--- a/crates/rune-modules/Cargo.toml
+++ b/crates/rune-modules/Cargo.toml
@@ -24,7 +24,7 @@ process = ["tokio/process"]
 signal = ["tokio/signal"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


killercup:
Wanted to test rune for a bit and saw it fail because OpenSSL isn't visible by default on NixOS. Since it's also always a pain to compile on macOS in my experience, I thought this might be a nice change. You lose the automatic security patches from Linux distributions, but gain a little more sanity by using an audited Rust implementation instead.

And having gzip compression for HTTP is a just a good thing.

shekohex (approved):

udoprog:
LGTM, thank you!

PR #57, opened 2020-09-12
Add https::Response.json

killercup:
Uses reqwest's JSON deserialization to create a `runestick::Value` without reading the response as string first. This also fixes the `http.rn` script which seems to have been broken by recent additions of `?` and `await`.


.github/workflows/build.yml
```diff
@@ -22,3 +22,4 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ steps.rust_toolchain.outputs.rustc_hash }}
       - run: cargo test --all
+      - run: cargo run --bin rune -- scripts/*
```
udoprog, line 25: This won't run every script. It will currently only pick up the last one:

https://github.com/rune-rs/rune/blob/master/crates/rune-cli/src/main.rs#L141

killercup: Ah, that's too bad. I'll make it a proper loop.

udoprog: I also realized that we don't set an exit status in case the script errors :)

killercup: Just saw that too! Anyway, updated now with a fancy bash for-loop; hope that works on CI. Let's do further fixes for that in a future PR?

.github/workflows/build.yml
```diff
@@ -21,4 +21,7 @@ jobs:
             ~/.cargo/bin
             target
           key: ${{ runner.os }}-cargo-${{ steps.rust_toolchain.outputs.rustc_hash }}
-      - run: cargo test --all
+      - name: Run tests
+        run: cargo test --all
+      - name: Execute example scripts
+        run: for script in scripts/*.rn; do cargo run --bin rune -- --experimental -O macros=true $script; done
```
udoprog, line 27: Feel free to unroll the loop (#59) :)

killercup: done + rebased

PR #60, opened 2020-09-12
Some VSCode extension fixes

killercup:


editors/code/src/main.ts
```diff
@@ -62,11 +66,19 @@ async function tryActivate(_context: vscode.ExtensionContext) {
...
-        return path.join(process.env.RUNE_DEBUG_FOLDER, `rune-languageserver${platform.ext}`);
+        const p = path.join(process.env.RUNE_DEBUG_FOLDER, exe);
+        if (existsSync(p)) { return p; }
+        else {
+            log.warn('env var `RUNE_DEBUG_FOLDER` is set but', p, 'does not exist');
+        }
     }
 
-    console.debug("Cannot find a command for the Rune Language Server.");
+    const p = path.join(process.env.HOME || '~', '.cargo', 'bin', exe);
```
udoprog, line 78: Neat! Possibly use [`os.homedir()`](https://nodejs.org/api/os.html#os_os_homedir) instead? I don't believe tilde expansion works for node on Windows unfortunately, and it correctly resolves to `USERPROFILE` where cargo home is by default on Windows.

editors/code/src/util.ts
```diff
@@ -0,0 +1,51 @@
+import * as vscode from 'vscode';
+import { inspect } from "util";
+
+// adapted from https://github.com/rust-analyzer/rust-analyzer/blob/ab432b36de5d6a370dbaad923f9a475a00fbf220/editors/code/src/util.ts#L16
```
udoprog, line 4: Very nice steal! The default `console.log` doesn't tell you the source of the log apart from the base name :)

crates/runestick-macros/src/context.rs
```diff
@@ -102,6 +102,7 @@ impl Context {
         let output = FieldAttrs::default();
 
         for attr in attrs {
+            #[allow(clippy::never_loop)] // I guess this is on purpose?
```
udoprog, line 105: Yep, meta parsing was removed. But will most likely be added back in again as more things are needed.

udoprog (approved):

udoprog:
Unrelated failure (which will be fixed!)

Issue #43, opened 2020-09-09
Support attributes


udoprog:
Support rust style attributes;

```rust
#[hello]
fn test() {
}
```

* [ ] Implement parsing.
* [ ] Add compiler support to process attributes.

killercup:
FYI, this is one of the main things I dislike in Rust's syntax because of the typing overhead and the fact that pretty much any tokens are valid in `#[here]`. Using `@Foo(param=foo)` and supporting exactly this style of calling a macro would reduce complexity quite a bit.

udoprog:
> FYI, this is one of the main things I dislike in Rust's syntax because of the typing overhead and the fact that pretty much any tokens are valid in `#[here]`. Using `@Foo(param=foo)` (or `#[Foo(param=foo)]` if you like typing braces) and supporting exactly this style of calling a macro would reduce complexity quite a bit.

I appreciate the point of view. But this is another instance when I want to maintain feature parity with Rust to make it possible to more easily transfer ideas. A note is that attributes are pretty much macro calls with different call syntax (and are associated with the AST they are tacked on to). You can see this in `syn`, with the [`Macro`](https://docs.rs/syn/1/syn/struct.Macro.html) (macro call) and [`Attribute`](https://docs.rs/syn/1/syn/struct.Attribute.html). They both take a `Path` and a raw `TokenStream`.

Alongside this, `syn` also provides a convenience function to [parse its content as meta items](https://docs.rs/syn/1/syn/struct.Attribute.html#method.parse_meta). Something like this can be done to unobtrusively enforce more opinion. Or something like [`darling`](https://github.com/TedDriggs/darling) can be ported to Rune.

ModProg:
Might look into this in the next few days, what should be the input of a function handling a macro? everything after the path? or should it only be the "input" as it is in rust: https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros i.e. if you call #[attribute(some_input)] the macro implementation would only get `some_input` and not  `(some_input)`.

ModProg:
this would mean, that a macro could not detect whether they got `#[path(...)]`, `#[path[...]]`, `#[path={...}` or `#[path = ...]` . Though I guess that should be fine, as it would be strange to have a different behavior depending on the exact syntax used to invoke (similar to how normal macros also don't differentiate`.

udoprog:
[...]

PR #83, opened 2020-09-19
Parse Attributes and attach them to their respective AST elements

dillonhicks:
**Issue:** #43

This PR enables parsing of attributes in various context and
attaching the attributes to their respective AST nodes. Attributes closely
follow grammar set by Rust and `ast::Attribute` elements are modeled
after `syn::Attribute`. The biggest difference is that Rune attributes
accept a wider range of literals for literal assignment, specifically :

`#[x = #{"yes": "it works"}]`

This PR **DOES NOT**:

* Provide parsing for the `ast::Attribute#input` TokenStream as a Meta
  item like syn::Attribute.
* Add the compiler support to rewrite `///` and `//!` comments as
  their respective `#[doc = ...]` & `#![doc = ...]` attributes.
* Provide compiler functionality for handling ast::Attribute elements
  in any meaningful way.  Attributes will be parsed if in a
  syntactically correct position but they will be ignored siltently.

See:

* [Rust Attribute Refrence](https://doc.rust-lang.org/reference/attributes.html)
* [syn::Attribute](https://docs.rs/syn/1.0.41/syn/struct.Attribute.html)
* *Demo*: ./scripts/attributes.rn


udoprog:
Thanks for the PR!

There were a couple of things I wanted to change, so I've rebased and modified the PR with the following:
* Associated attributes with `BlockExpr` and `AsyncExpr` expressions (instead of `Block`).
* Adjusted spans where attributes were added so that they include their attributes.
* Simplified parsing of attributes.
* Rejected attributes at the indexing stage (this will be replaced with the actual macro calls once attributes are actually supported beyond the parsing stage). We don't want programs with attributes to compile just yet, but soon ;).

Since you went through the work of adding `ast::Lit` (:rocket:), I've also switched everything over to use this as the representation for a literal by introducing `ast::ExprLit`.

dillonhicks:
> Thanks for the PR!
> 
> There were a couple of things I wanted to change, so I've rebased and modified the PR with the following:

I am glad to help! This was a bigger change than I expected it to be, so sorry about that. I would have been happy to make the changes you wanted, if even to just help familiarize myself with the code a bit more. But done is done 🕺 .

>     * Associated attributes with `BlockExpr` and `AsyncExpr` expressions (instead of `Block`).

Cool. I agree that pushing the attributes down into the block was less than ideal.

>     * Adjusted spans where attributes were added so that they include their attributes.
>
>     * Simplified parsing of attributes.

👍 

>     * Rejected attributes at the indexing stage (this will be replaced with the actual macro calls once attributes are actually supported beyond the parsing stage). We don't want programs with attributes to compile just yet, but soon ;).

The good ol' 🐔 and the 🥚  problem, I want to run a program with an unfinished feature (attributes) but that feature require the feature itself to enable the desired behavior 🤯 (I'm assuming in the future there will be a `#![feature(<feature_name>)]`)  for unstable features?). I guess I could have added a `-Oattributes` but I am not comfortable enough with the compiler yet to modify it with any confidence. I'll am going to glean some insight from looking your changes in a bit.

> 
> Since you went through the work of adding `ast::Lit` (🚀), I've also switched everything over to use this as the representation for a literal by introducing `ast::ExprLit` (to match [`syn::ExprLit`](https://docs.rs/syn/1/syn/struct.ExprLit.html)).

😎 I thought about making that change -- but this change was large enough already I felt a bit sheepish about "fixing" things that were not broken in this pass. 

On to the next task 🚀 !

PR #84, opened 2020-09-19
fix some broken links to the book in the README

dillonhicks:
Links in the README were throwing 404s. They should now point to the correct urls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
     <b>Visit the site 🌐</b>
 </a>
 -
-<a href="https://rune-rs.github.io/bool/">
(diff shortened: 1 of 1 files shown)


udoprog:
Nice, thank you!

This change should actually go into `crates/rune/src/lib.rs`, and we then use [`tools/readmes.rn`](https://github.com/rune-rs/rune/blob/master/tools/readmes.rn) to regenerate the readmes.

The links are actually wrong everywhere. A search in visual studio code shows all the places:
image

dillonhicks:
Cool, thanks for the heads up about the `tools/readme.rn`.

PR #92, opened 2020-09-21
Add Parsing of Visibility Levels for Items and Fields

dillonhicks:
Parsing support for Visibility levels as part of supporting modules (#5).

#### Examples

This is now accepted by the parser but rejected by the compiler: 

```rust 

pub use foo;

pub(crate) struct Bar {
    pub(super) baz,
}

impl Bar {
    pub(in crate::xyz) fn do_stuff() {
       // ... 
     }
}

```

#### Changes

* Adds an `ast::Visibility` which is a subset of `syn::Visibility` to parse `pub` and `pub(...)`. 
* Adds a `pub visibility: Option<ast::Visibility>` to:
  * `StructItem`
  * `Field`
  * `EnumItem`
  * `FnItem`
  * `UseItem`
  * `ModItem`
* `Ident`: Add `From<T>` impls for keywords that can be `Idents` in some contexts. 
   * Added the impls to match `syn::Ident` but specifically needed for `ast::Crate` and `ast::Super`, see `Path` changes.
* `Path`: Updates for support `pub(in path)`,`pub(crate)`, and `pub(super)`
    * Allow optional leading `::`  (rust-likeness)
    * Allow `crate` and `super` to be in the first position of a path.
* Add a few new tokens for keywords
* Reject usage of visibility levels in the compiler with an internal error.


crates/rune/src/ast/item.rs
```diff
@@ -81,7 +88,14 @@ impl Item {
 
 impl Peek for Item {
     fn peek(t1: Option<ast::Token>, t2: Option<ast::Token>) -> bool {
```
udoprog, line 90: These peek impls should probably be demoted to not be implementations of `Peek`, and just associated method (e.g. remove this and just rely on `peek_as_stmt`). Because they won't currently match items with attributes which will mess with parsing in macros.

crates/rune/src/ast/path.rs
```diff
@@ -49,6 +52,55 @@ impl Peek for Path {
...
+/// parse_all::<ast::Path>("::ab::cd").unwrap();
+/// parse_all::<ast::Path>("crate").unwrap();
+/// parse_all::<ast::Path>("super").unwrap();
+/// parse_all::<ast::Path>("crate::foo").unwrap();
+/// parse_all::<ast::Path>("super::bar").unwrap();
+/// parse_all::<ast::Path>("::super").unwrap();
+/// parse_all::<ast::Path>("::crate").unwrap();
+/// ```
+///
+impl Parse for Path {
```
udoprog, line 74: IIUC, adding a `PathSegment` enum (as mentioned earlier) would allow for using a generated `Parse` impl again.

crates/rune/src/ast/vis.rs
```diff
@@ -0,0 +1,80 @@
+use crate::ast;
+use crate::{Parse, ParseError, ParseErrorKind, Parser, Peek, Spanned, ToTokens};
+
+/// Visibility level restricted to some path: pub(self) or pub(super) or pub(crate) or pub(in some::module).
+#[derive(Debug, Clone, ToTokens, Spanned)]
+pub enum Visibility {
```
udoprog, line 6: I'm considering if instead of using `Option<Visibility>` everywhere, we could use [`Visibility::Inherited` as `syn` does](https://docs.rs/syn/1/syn/enum.Visibility.html). The only difference here is that you'd have to derive `OptionSpanned`. But it seems reasonable since this is also how the [`vis` metavariable in proper Rust macros work](https://doc.rust-lang.org/reference/macros-by-example.html#metavariables) (they can parse nothing).

udoprog (changes requested):
Looks really good. The only thing I really would insist on changing is getting rid of the `From<T>` impls for `Ident`. They are potential implementation hazards that can be mostly avoided with a `PathSegment` enum. Which is a deviation from `syn`, but a good one for this project right now.

crates/rune/src/ast/ident.rs
```diff
@@ -63,3 +63,39 @@ impl<'a> Resolve<'a> for Ident {
...
+}
+
+impl std::convert::From<ast::Extern> for Ident {
+    fn from(extern_: ast::Extern) -> Self {
+        Ident {
+            token: extern_.token,
+            kind: ast::StringSource::Text,
+        }
+    }
+}
```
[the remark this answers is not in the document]
dillonhicks: Sounds good to me, I started with something similar but I switched because "it's what syn does."

crates/rune/src/ast/vis.rs
dillonhicks: Neat, I'll check out OptionSpanned.

dillonhicks: I'll check out OptionSpanned. My first attempt to include `Inherited` was to extend the `Spanned` macro to just return `Default::default()` on Unit variants of enums but that seemed to break things rather quickly. I didn't investigate it too much because I had questions "how does one peek a tokenless item?" and  Option<T> seemed to be the closest mechanism to handle tokenless "meta" items.

udoprog: Ok!

For reference, this is how it's done:
* Processing a field that implements `OptionSpanned`: https://github.com/rune-rs/rune/blob/08f2633e11e03d8bca450b6bcd2de3a13c9be2a9/crates/rune/src/ast/item_enum.rs#L90
* Right below it, the thing which derives `OptionSpanned`: https://github.com/rune-rs/rune/blob/08f2633e11e03d8bca450b6bcd2de3a13c9be2a9/crates/rune/src/ast/item_enum.rs#L98

dillonhicks: Sweet, thanks!

I assume that the OptionSpanned macro enforces a single Unit variant and makes it the default.

``` 
    /// An empty enum body.
    EmptyBody, 
```

Would it make sense to make this obvious with a `#[rune(default)]` ?

```
    /// An empty enum body.
    #[rune(default)]
    EmptyBody,
```

udoprog: It doesn't enforce a single unit variant, but unit variants always produce `None` in their `fn option_span(&self)` impl. If there's no unit variant, the struct can derive `Spanned`.

`OptionSpanned` is necessary if:
* There's a unit variant.
* All fields could potentially produce no span information (not sure this case is handled correctly right now though!).

dillonhicks:
After 844ee31e75a5f9f790c45b19f435ba0e39026911, I just need to fix up the "demotion of peek impls for items" https://github.com/rune-rs/rune/pull/92#discussion_r491765255 . I should get to that tomorrow.

crates/rune-macros/src/parse.rs
```diff
@@ -95,28 +95,60 @@ impl Expander {
...
+        let ast_attributes_ident: syn::Ident = syn::parse2(quote! {attributes})
+            .expect("Parse proc_macro could not create needed resource Ident(\"attributes\")");
+
+        let mut has_ast_attributes = false;
         let mut fields = Vec::new();
 
         for field in &named.named {
             let _ = self.ctx.parse_field_attributes(&field.attrs)?;
             let ident = self.ctx.field_ident(&field)?;
+            if ident.eq(&ast_attributes_ident) {
```
udoprog, line 107: Nice. I like that you're generating `parse_with_attributes` functions since they all follow the same pattern. Hopefully we can do more of that!

Could you make this optional behavior activated by an attribute `#[rune(attributes)]` instead? I prefer the explicitness of it over defaulting to the name. You can see [how they are parsed here](https://github.com/rune-rs/rune/blob/master/crates/rune-macros/src/context.rs#L134).

As for synthesizing identifiers (the parse/quote thing), check out the [`internals` module](https://github.com/rune-rs/rune/blob/master/crates/rune-macros/src/internals.rs#L8). A `Symbol` can be used for that (it implements `ToTokens` for this purpose).

dillonhicks: Can do. After making the "parse -> parse_with_attributes"  pattern few times I thought I'd give generating it a go.

crates/rune/src/index.rs
```diff
@@ -424,6 +436,12 @@ impl Index<ast::Expr> for Indexer<'_> {
     fn index(&mut self, expr: &ast::Expr) -> CompileResult<()> {
         let span = expr.span();
         log::trace!("Expr => {:?}", self.source.source(span));
+        if expr.has_attributes() {
```
udoprog, line 439: This is a pretty fine solution for now, but will be pretty short lived once we start gradually introducing attribute semantics to various types 🙃. Maybe rename it `has_unsupported_attributes` or some such and we can gradually remove variants from it once they gain support in the compiler.

crates/rune/src/ast/path.rs
```diff
@@ -68,3 +72,68 @@ impl<'a> Resolve<'a> for Path {
...
+    }
+}
+
+impl Peek for PathSegment {
+    fn peek(t1: Option<ast::Token>, _t2: Option<ast::Token>) -> bool {
+        matches!(peek!(t1).kind, ast::Kind::Ident(_) | ast::Kind::Crate | ast::Kind::Super)
+    }
+}
+
+impl<'a> Resolve<'a> for PathSegment {
```
udoprog, line 123: The urge to make this shortcut is strong, but I don't think we should let PathSegment's resolve to strings regardless. I.e. we never want `super` or `crate` to become part of an `Item`. What they do is structural: they change the resolution to the crate root, or move "up" one level when resolving paths.

To see why, consider what would happen here (where `into_components()` is used) if the path contains `super` or `crate`: https://github.com/rune-rs/rune/blob/4d4bb71ba7df2938f60298b743ddd58ccea672a6/crates/rune/src/index.rs#L729

crates/rune/src/ast/item.rs
```diff
@@ -47,12 +67,21 @@ impl Item {
...
+    /// Parse an Item attaching the given Attributes
+    pub fn parse_with_attributes(
+        parser: &mut Parser,
+        attributes: Vec<ast::Attribute>,
+    ) -> Result<Self, ParseError> {
         let t = parser.token_peek_eof()?;
 
-        Ok(match t.kind {
+        let kind = if t.kind == ast::Kind::Pub {
+            let t2 = parser.token_peek2_eof()?;
```
udoprog, line 78: This probably doesn't work for restricted visibility (e.g. `pub(crate)`)? So we'll need something `parse_with_attributes_and_vis` (or `parse_with_meta`) for the items to do this correctly.

I don't have too much of an issue about it for now, since visibility is not supported anyway. But should be fixed soon:ish.

udoprog (approved):
One minor thing that can be fixed in a future PR, but I'm ready to merge now. Thank you!

I especially appreciate the amount of parse tests added. They really help!

crates/rune/src/ast/item.rs
dillonhicks: Good catch. Yeah it will definitely fail soon as there is an actual use/test of it.

Makes me wonder if we can come up with general pattern for common "prefixes" of ast structs?

PR #93, opened 2020-09-21
Initial support for const exprs

udoprog:
This introduces the basic plumbing for constant expressions. ~~We're still missing recursive evaluations, e.g. referencing other constant values and building a single `ConstValue` out of it.~~

But this performs the basics of plugging constant expressions into the build queue which handles things like unused diagnostics.

Added cycle checking and a number of operations.

#### Added operations

* Resolve other constant values in scope.
* Integer operations:
  * `+`, `-`, `*`, `/`.
  * `<<`, `>>`.
  * `<`, `<=`, `==`, `>=`, `>` (producing boolean).
* Float operations:
  * `+`, `-`, `*`, `/`.
  * `<`, `<=`, `==`, `>=`, `>` (producing boolean).
* String templates with constants interpolated.
* Define a constant boolean.
* While loops with expression-like conditions `while <expr> { <block> }`.
* If expressions with expression-like conditions `if <expr> { <then> } [else if { <else-if> }]* [else { <else> }]`.

Also added an [evaluation budget](https://github.com/rune-rs/rune/pull/93/files#diff-a7a441d34213eeea97100238a1eb797dR457) which defaults to `1000`. This prevents infinite computations like `while true {}` from never completing.

There's lots of things missing, the [constant compiler](https://github.com/rune-rs/rune/pull/93/files#diff-a56aa8c677d5bed92e6612b8acf46ec4R28) can (and should) be expanded to support more-or-less all language features which does not have side effects. But since it's a separate execution model it will always mean duplicating work.


dillonhicks (commented):
This PR is insightful. It'd make a good example for "_How to add support for a feature_" in a Dev-Handbook or Wiki.

udoprog:
@dillonhicks Thanks :). I'm afraid this might end up being a bit too complicated to be a nice streamlined intro.

PR #97, opened 2020-09-22
Introduce Vm::send_execute to allow for threaded async execution

udoprog:
Introduces `Vm::send_execute`, which allows for performing the execution on a thread pool like [tokio::spawn](https://docs.rs/tokio/0/tokio/runtime/struct.Runtime.html#method.spawn).

The fundamental problem is the `Vm` contains the equivalent of a stack of `Rc<RefCell<T>>`, meaning it's aggressively `!Send`. To rationalize this, consider what would happen if we cloned a Vm with its stack and sent it to a different thread. The non-atomic refcount of any values which escaped to the other thread could be mutated by multiple threads and bad things would happen.

What this patch allows for is tying the execution of a `Vm` to a single future through `VmSendExecution::async_complete`. The future itself guarantees that the `Vm` will only be polled by one thread at a time due to its convenient [`self: Pin<&mut Self>` signature](https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll). This also means that while the `Vm` can be sent to a different thread between polls, all of its corresponding state will only be accessed by one thread at a time. There's no "racy access" of its interior state or refcounts*.

> *: whether this in itself is a strong enough guarantee that no data races will happen I'm not entirely sure of. It depends a bit on how multithreaded access over raw pointer behaves and needs to be investigated more. Luckily we control the implementation of `Shared<T>`, and can hopefully fix it if needed.

So we wrap the execution in a `VmSendExecution`, which effectively hides any functions which would otherwise allow values escaping from the execution in a way to send a clone.

#### Rejecting non-Send arguments to `send_execute`

The `args` argument to `send_execute` has a `Send` bound. This means that arguments which are `!Send` cannot be smuggled into the stack, like `Value`. The following program doesn't compile, because it tries to encode a `Value`, even if its `Value::Integer` which strictly speaking is a `Send` variant, it's still rejected on a type level because `Value` is `!Send` due to its other variants:

```rust
#[tokio::main]
async fn main() -> runestick::Result<()> {
    let (context, unit) = rune::testing::build(
        r#"
        async fn main(timeout) {
            time::delay_for(time::Duration::from_secs(timeout)).await
        }
        "#,
    )?;

    let main = runestick::Hash::type_hash(&["main"]);

    let vm = runestick::Vm::new(context.clone(), unit.clone());

    let execution = vm.clone().send_execute(main, (runestick::Value::Integer(5),))?; // <- here

    let t = tokio::spawn(async move {
        execution.async_complete().await.unwrap();
        println!("timer ticked");
    });

    t.await.unwrap()
}
```

> Note: this blows up so spectacularly that it [requires a gist to show](https://gist.github.com/udoprog/02d0c4ec916f281f13b12183f01a1334).

We can provide the number `5` which *is* `Send`. Internally it will still be converted into a `Value` when encoded onto the stack. But we have now guaranteed that the stack has exclusive ownership of its values.

```rust
#[tokio::main]
async fn main() -> runestick::Result<()> {
    // .. trimmed

    let execution = vm.clone().send_execute(main, (5,))?; // <- here

    let t = tokio::spawn(async move {
        execution.async_complete().await.unwrap();
        println!("timer ticked");
    });

    t.await.unwrap()
}
```
Finally, `async_complete` takes a `mut self`, effectively consuming all `Vm` state *except* for the `Value` returned. Which is now safely stranded in the thread that left it because it is `!Send`.


dillonhicks:
> The future itself guarantees that the Vm will only be polled by one thread at a time due to its convenient self: Pin<&mut Self> signature. This also means that while the Vm can be sent to a different thread between polls, all of its corresponding state will only be accessed by one thread at a time. There's no "racy access" of its interior state or refcounts*.
> 
>     *: whether this in itself is a strong enough guarantee that no data races will happen I'm not entirely sure of. It depends a bit on how multithreaded access over raw pointer behaves and needs to be investigated more. Luckily we control the implementation of Shared<T>, and can hopefully fix it if needed.
> 

Can you clarify multithreaded access in this sentence? " It depends a bit on how multithreaded access over raw pointer behaves and needs to be investigated more."

Do you mean multithreaded access as in "during the lifetime of the pointer, more than 1 unique thread will ever access the raw pointer, but each access is guaranteed to be mutually exclusive"?

udoprog:
> Do you mean multithreaded access as in "during the lifetime of the pointer, more than 1 unique thread will ever access the raw pointer, but each access is guaranteed to be mutually exclusive"?

Yes. This. It's guaranteed by the signature of `Future::poll`. Any data stored in the future can only be accessed by one thread at a time. Because in order to poll it you need to have a pinned *exclusive* reference. The VM execution is stored in the future, and we (hopefully) encapsulate it in a way which prevents it or any data related to it to leak while it's being polled in safe code.

Issue #73, opened 2020-09-16
Define a format for external definitions


udoprog:
This will be required for native items (functions, types, macros, ...) to be known by the language server so that they can compile without having to load the native modules in.

Preliminarily the definition format can be like Rune, but without declaration bodies:

```rust
#[definition(module = "game")]

/// Always opaque, body is not known.
struct Player {
    /// The health of the player.
    #[getter]
    health,
}

impl Npc {
    /// The documentation for an instance function.
    fn instance_fn();
}

/// The documentation for a free function.
fn free_fn();
```

Furthermore, we can load the definition into the module, which will then fail to load in case definitions are missing or wrong:

```rust
fn module() -> Result<runestick::Module, runestick::ContextError> {
    let module = runestick::Module::from_definition(include_str!("path/to/module.d.rs"))?;
    // install things into the module.
    module
}
```

PR #100, opened 2020-09-22
Add the `rand` module

aspenluxxxy:
Adds the `rand` module.

Exposes two types: `rand::WyRand` and `rand::Pcg64`, which are RNGs. Each has the following functions: `int() -> Integer` and `int_range(lower: Integer, upper: Integer) -> Integer`

In addition, `rand::int[_range]` exposes a single-use WyRand generator


udoprog (changes requested):
Module looks great. Could you remove and `.gitignore` the `.idea` files?

My impression of IntelliJ is that it does just fine importing projects. And I won't be able to keep them up-to-date seeing as I don't use it. It's different with e.g. `.vscode/launch.json`, since this is just a generic launch configuration. But an incomplete `.idea` project listing seems like it could cause editor confusion!

udoprog:
Also looking forward to #73, so we can get somewhere to put api docs :).

crates/rune-modules/src/rand.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Create a new RNG instance, using a custom seed.
+    fn new_seed(seed: i64) -> Self {
+        Self {
+            inner: nanorand::Pcg64::new_seed(seed as u128),
+        }
+    }
+
+    /// Generate a random integer
+    fn int(&mut self) -> Value {
```
udoprog, line 113: FTR, these can simply return `i64`. Conversion can be handled by Rune (using the `ToValue` trait). Returning a `Value` doesn't hurt though.

udoprog (approved):

udoprog:
Thank you!

PR #125, opened 2020-09-29
Request for advice for a way forward for resolving Self and self

dillonhicks:
This isn't rebased on the current changes that support "path tree" and is I have crate and super working reliably but Self and self are problematic. 

So I need help tracking down these two problems:

* Resolving`Self` works in the simplest cases but fails in nested impl contexts, see: https://github.com/rune-rs/rune/compare/master...dillonhicks:keyword-paths-advice-pr?expand=1#diff-18418503cbb8b7076324353e3c9c592a . It looks like it indexes the inner function `Foo::outer_1::$block0::Bar::inner_2` but the later query produces no meta. Any pointers?

* Resolving `self` fails if not prefixed with `::`  because it is parsed as a `Expr::Self_`. Should I just remove this in favor of parsing "paths"?


udoprog:
> Resolving self fails if not prefixed with :: because it is parsed as a Expr::Self_. Should I just remove this in favor of parsing "paths"?

Yeah. Definitely.

> Resolving Self works in the simplest cases but fails in nested impl contexts, see: [..]

It should be possible to resolve `Self` from `impl_items` for now: https://github.com/rune-rs/rune/blob/d8caf6143992ad60f282a66d163763d9618c5850/crates/rune/src/indexing/index.rs#L44

You'll have to store this in `Query` and look it up through an opaque identifier. Similarly to how it's done to figure out where a path is resolved from today. It follows that of `Self` is encountered outside of an impl block it should cause an error.

dillonhicks:
I rebased onto 8a43abb. I'm going to give this one last attempt before I admit defeat. 

I am going to focus on `Self` and ignore `self` at the moment since that might be a bit of a distraction because of the parsing changes.

Thanks!

PR #128, opened 2020-09-29
Add parsing for Self and self and implement item resolution

udoprog:
Adapted the parsing from #125, and the first step to get things up and running without stepping on each others toes too much.


udoprog:
So I was poking at the resolution problem and noticed that there were a large number of things that needed to be changed. So I've implemented those changes here as well, since they were nontrivial.

PR #147, opened 2020-10-10
Implement Args for Vec

MinusGix:
The first commit changes the `Args::count` method to take a self parameter, as the size for a Vector can not be known at  compile-time like the size of tuples. Hopefully, such a change won't cause worse optimization for tuples as all `Args` are used as generic parameters and so are known what they are compile-time.  
  
The second commit implements Args for a `Vec<runestick::Value>`, as well as adding an example of it being used. The use-case for this is when you want to pass parameters of an unknown amount into a rune function.  
This could be changed to be more generic instead, such as implementing `Args` for `Vec<T> where T: ToValue`. That would include `Value`, and other Vectors of types which could be transformed into values. If you'd prefer, I could change the pr to implement it for `T: ToValue`.

### DIFF
--- a/crates/rune/examples/vec_args.rs
+++ b/crates/rune/examples/vec_args.rs
@@ -1,0 +1,42 @@
+use std::sync::Arc;
+
+use rune::{Errors, Options, Sources, Warnings};
+use runestick::{Context, Module, Source, Value, Vm, VmError};
+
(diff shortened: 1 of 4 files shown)


udoprog:
This is perfect. Much appreciated!

PR #151, opened 2020-10-11
Export SyncFunction

MinusGix:
This exports the newly created `SyncFunction` so that it is publicly available.


udoprog:
Thanks :)

PR #168, opened 2020-11-23
add `file!()` and `line!()` macros

tgolsson:
Hello!

From discussion on Discord - this adds the `file!()` and `line!()` macros, which transform into string and number literals representing the current filename and line number (1-indexed). I tried to follow the pattern from `format!()`, but it seems a bit awkward being unable to implement these in runestick, instead feature-gating it in `rune-modules`.

```
➜ cat test.rn
pub fn main() {
   println(`Hello from ${file!()}:${line!()}`);
}

➜ target/release/rune test.rn
Hello from test.rn:2
== () (25.346µs)
```

### DIFF
--- a/crates/rune-modules/src/macros.rs
+++ b/crates/rune-modules/src/macros.rs
@@ -15,9 +15,17 @@
 //! ```rust
 //! # fn main() -> runestick::Result<()> {
 //! let mut context = runestick::Context::with_default_modules()?;
-//! context.install(&rune_modules::macros::module()?)?;
+//! context.install(&rune_modules::macros::module(true)?)?;
 //! # Ok(())
 //! # }
+//! ```
+//!
+//! Use it in Rune:
+//!
+//! ```rust,ignore
+//! fn main() {
+//!     println(`Hello from ${file!()}:${line!()});
+//! }
 //! ```
 
 use rune::{Parser, TokenStream};
--- a/crates/rune-modules/src/macros.rs
+++ b/crates/rune-modules/src/macros.rs
@@ -1,0 +1,57 @@
+//! `std::macros` module for the [Rune Language].
+//!
+//! [Rune Language]: https://rune-rs.github.io
+//!
+//! ## Usage
+//!
+//! Add the following to your `Cargo.toml`:
+//!
+//! ```toml
+//! rune-modules = {version = "0.7.0", features = ["macros"]}
+//! ```
+//!
+//! Install it into your context:
+//!
+//! ```rust
+//! # fn main() -> runestick::Result<()> {
+//! let mut context = runestick::Context::with_default_modules()?;
+//! context.install(&rune_modules::macros::module()?)?;
+//! # Ok(())
+//! # }
+//! ```
+
+use rune::{Parser, TokenStream};
+
+/// Construct the supplemental `std::macros` module.
+pub fn module(_unused: bool) -> Result<runestick::Module, runestick::ContextError> {
+    let mut builtins = runestick::Module::new(&["std", "macros", "builtin"]);
+    builtins.macro_(&["file"], emit_file)?;
+    builtins.macro_(&["line"], emit_line)?;
+    Ok(builtins)
+}
+
+/// Implementation for the `line!()` macro
+pub(crate) fn emit_line(stream: &TokenStream) -> runestick::Result<TokenStream> {
+    let mut parser = Parser::from_token_stream(stream);
+
+    parser.eof()?;
+
+    Ok(rune::quote!(
+        #[builtin]
+        line!()
+    )
+    .into_token_stream())
+}
+
+/// Implementation for the `file!()` macro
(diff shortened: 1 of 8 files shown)


crates/rune/src/indexing/index.rs
```diff
@@ -285,6 +289,46 @@ impl<'a> Indexer<'a> {
...
+            span: ast.span(),
+            value: node,
+        }))
+    }
+
+    /// Expand a macro returning the current line for where the macro invocation begins
+    fn expand_line_macro(&mut self, ast: &mut ast::MacroCall) -> Result<BuiltInMacro, ParseError> {
+        let (l, _) = self
+            .source
+            .position_to_utf16cu_line_char(ast.open.span.start.into_usize())
```
udoprog, line 313: hah, not that it matters (line count should be the same), but it would be nice with an utf8 version of this as well.

crates/rune/src/indexing/index.rs
```diff
@@ -285,6 +289,46 @@ impl<'a> Indexer<'a> {
         }))
     }
 
+    /// Expand a macro returning the current file
+    fn expand_file_macro(&mut self, ast: &mut ast::MacroCall) -> Result<BuiltInMacro, ParseError> {
+        let id = self.storage.insert_str(self.source.name());
```
udoprog, line 294: Rune sources can be completely in-memory, and names do not strictly speaking need to match the path. The best approach here is probably to use `self.source.path()` and error in case the source doesn't have a path.

crates/rune-modules/src/macros.rs
```diff
@@ -0,0 +1,65 @@
...
+//! fn main() {
+//!     println(`Hello from ${file!()}:${line!()});
+//! }
+//! ```
+
+use rune::{Parser, TokenStream};
+
+/// Construct the supplemental `std::macros` module.
+pub fn module(_unused: bool) -> Result<runestick::Module, runestick::ContextError> {
+    let mut builtins = runestick::Module::new(&["std", "macros", "builtin"]);
```
udoprog, line 35: Guess this module is as good as any for now. Not sure where to put these macros since they just live immediately in `std` in Rust!

crates/rune/src/indexing/index.rs
udoprog: Also, principle this could use the `ast::Lit::new_with` constructor. But I only think it works for strings and a few more for now.

udoprog (commented):
Great job. Only the comment about using `self.source.path()` needs to be addressed for now. Especially nice touch re-using the assembly impl for `LitStr` and `LitNumber`!

crates/rune/src/indexing/index.rs
tgolsson: How'd that work - I can't see any information about the actual encoding of the incoming source? This should probably use the unicode variant though, not utf16.

crates/rune/src/indexing/index.rs
tgolsson: I'm not sure if that'd make sense as a user, though. I'd expect it to expand to *something* in all builds given that we organize code based on a filesystem hierarchy... I'd maybe argue that if anything, the fact that the location is a filesystem path is unnecessary after loading. It's just code from a named source whether that be a webpage, a ZIP archive or a file on disk.

crates/rune-modules/src/macros.rs
tgolsson: Yeah; It's a bit odd. They're actually (code organization wise) in std/core/macros, in a `mod builtin`. Since I'm adding them to the prelude I guess it's not too important where they live apart from requiring the correct rune-modules features?

tgolsson: Ah, this is a result of the `macro_export` causing macros to appear in the root. So I think this path is correct w.r.t. code organization but we don't have the same kludgy export mechanics that Rust has.

crates/rune/src/indexing/index.rs
udoprog: `utf16cu` stands for UTF-16 code units, the position function does some work to re-encode each character to utf-16 in order to find the column. An equivalent version for UTF-8 would simply rely on the emitted chars instead.

See: https://github.com/rune-rs/rune/blob/master/crates/runestick/src/source.rs#L135

Not a blocker though, since the line will be the same.

crates/rune-modules/src/macros.rs
udoprog: Ah, yeah. That seems fine to me then :)

crates/rune/src/indexing/index.rs
udoprog: All right. I'm fine either way really. Just rubs me the wrong way that `file!` does not mean a file semantically. But it's not a big deal since I expect it to only be used diagnostically!

udoprog:
All blockers are resolved. :rocket:

PR #176, opened 2020-11-28
make ContextSignature, ContextTypeInfo public

tgolsson:
These are part of the public Context API but the enum is unusable since it's not exported. And also, since they are not public you don't get proper information shown on docs.rs:
image


udoprog:
Thanks!

PR #178, opened 2020-11-29
Add std::any module and clear up some function type hash confusion

tgolsson:
I think this is it!  Took a bunch of iterations but I think the overall structure is good. No other constants supported than type name but it felt quite nice to implement in the end. Not sure if there are more cases I should test?


crates/runestick/src/interface.rs
```diff
@@ -34,8 +34,35 @@ impl Interface {
...
+        let hash = Hash::instance_function(self.target.type_hash()?, Protocol::INTO_TYPE_NAME);
+        if let Some(name) = self.context.constant(hash) {
+            match name {
+                ConstValue::String(s) => return Ok(s.clone()),
+                ConstValue::StaticString(s) => return Ok((*s).to_string()),
+                _ => Err(VmError::expected::<String>(name.type_info()))?,
+            }
+        }
+
+        let hash = Hash::instance_function(self.target.type_hash()?, Protocol::INTO_TYPE_NAME);
```
udoprog, line 48: No need to calculate the hash twice 🙃

crates/runestick/src/function.rs
```diff
@@ -23,6 +23,7 @@ where
     Tuple: From<Box<[V]>>,
 {
     inner: Inner<V>,
+    hash: Hash,
```
udoprog, line 26: IIUC, Rtti for the other variants have access to the corresponding type hashes, so this probably only needs to be added to the `FnHandler` and `FnOffset` variants.

crates/runestick/src/interface.rs
```diff
@@ -34,8 +34,35 @@ impl Interface {
...
+        if let Some(name) = self.unit.constant(hash) {
+            match name {
+                ConstValue::String(s) => return Ok(s.clone()),
+                ConstValue::StaticString(s) => return Ok((*s).to_string()),
+                _ => Err(VmError::expected::<String>(name.type_info()))?,
+            }
```
udoprog, line 54: This can be replaced with:
```rust
return String::from_value(name);
```
See: https://github.com/rune-rs/rune/blob/main/crates/runestick/src/from_value.rs#L147

crates/runestick/src/modules/any.rs
```diff
@@ -0,0 +1,42 @@
+/*!
+
+*/
```
udoprog, line 3: rogue docs!

crates/runestick/tests/test_native_types.rs
```diff
@@ -0,0 +1,116 @@
+/*!
+Tests for `std::any::type_name_of_val(v)`
+*/
```
udoprog, line 3: Module-level documentation in Rust uses `//!`, like so:
```rust
//! Test for `std::any::type_name_of_val(v)`
```

crates/runestick/tests/test_native_types.rs
```diff
@@ -0,0 +1,116 @@
+/*!
```
udoprog, line 1: I'd suggest putting this and the other test case as a module under [`test_all` tests module in the `rune` crate](https://github.com/rune-rs/rune/tree/main/crates/rune/tests/test_all) instead. There's a couple of helpers you can use there, like the `rune!` macro to run rune code.

Note: also needs to be added to <https://github.com/rune-rs/rune/blob/main/crates/rune/tests/test_all/mod.rs> to run if there are new modules.

udoprog:
Good approach! Some minor comments.

crates/runestick/src/interface.rs
tgolsson: Unfortunately not, but maybe it should... This is ConstValue, not Value.

crates/runestick/src/function.rs
tgolsson: Fixed, thanks!

crates/runestick/tests/test_native_types.rs
tgolsson: Will do! :-)

crates/runestick/src/interface.rs
udoprog: Aha. Right :)

crates/runestick/tests/test_native_types.rs
tgolsson: Done!

tgolsson:
@udoprog Thanks for the feedback, I think all points are addressed now.

udoprog:
Really nice. Thank you!

PR #185, opened 2020-12-01
expand on Result, Option, Iterator

tgolsson:
Some missing ones I found useful!

* Result::{expect, unwrap, map, and_then}
* Option::{expect, unwrap, map, and_then}
* Iterator::{sum, product, fold}
* Tests


udoprog:
Interesting approach inspecting the first value in the iterator. Not sure that's where we want to end up, but it's certainly better than type-bespoke functions (e.g. `sum_int`) for now.

PR #186, opened 2020-12-01
minimal support for non-default LSPs

tgolsson:
This adds support for using the LSP implemenation from other workspaces to expose custom bindings without reimplementing the whole binary. To make this easier, it also adds a VSCode extension setting so you can set the binary path to where your custom LSP lives. This gives you a workspace file with a minimum:

```json
{
    "rune.binaryPath": "/home/tgolsson/Repos/rune/target/release/rune-languageserver"
}
```

This is me using a locally built version of the rune languageserver as I build the whole toolchain locally.

For a fully custom integration, I do this when I'm working on my game-engine:

```rust
extern crate binds;
fn main() -> anyhow::Result<()> {
        let mut context = rune_modules::default_context()?;

        for module in binds::all_modules() {
            context.install(&module)?;
        }

        let mut options = rune::Options::default();
        options.macros(true);

        rune_languageserver::run(context, options)
}
```

and use a workspace file targeting this binary instead.

### DIFF
--- a/crates/rune-languageserver/src/lib.rs
+++ b/crates/rune-languageserver/src/lib.rs
@@ -55,3 +55,182 @@ pub use crate::connection::stdio;
 pub use crate::connection::{Input, Output};
 pub use crate::server::Server;
 pub use crate::state::State;
+use anyhow::Result;
+use tokio::sync::mpsc;
+
+pub fn run(context: runestick::Context, options: rune::Options) -> Result<()> {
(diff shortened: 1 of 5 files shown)


udoprog (approved):
Great work. It's a good workaround for the lack of metadata from native modules. I'll take it for a spin as soon as I can to make sure everything works as expected, but it looks good.

PR #188, opened 2020-12-02
Added Pattern matching for True and False And a test.

genusistimelord:
Added the True and False Pattern for matching.


udoprog:
Thanks!

PR #193, opened 2020-12-03
Add fibonacci and brainfuck interpreter benchmarks

tgolsson:
As promised, I've added these two groups of benchmarks, along with some required utilities:

* Vec[...] operator 
* char::from_int converter
* Output redirect for tests
* A macro to create a vm with redirected output for tests

### DIFF
--- a/crates/runestick/src/modules/char.rs
+++ b/crates/runestick/src/modules/char.rs
@@ -11,14 +11,13 @@ pub fn module() -> Result<Module, ContextError> {
     Ok(module)
 }
 
-fn char_from_int_impl(value: Value) -> Result<Option<Value>, VmError> {
-    let inner: i64 = value.into_integer()?;
-    if inner < 0 {
+fn char_from_int_impl(value: i64) -> Result<Option<Value>, VmError> {
+    if value < 0 {
         Err(VmError::from(VmErrorKind::Underflow))
-    } else if inner > u32::MAX as i64 {
+    } else if value > u32::MAX as i64 {
         Err(VmError::from(VmErrorKind::Overflow))
     } else {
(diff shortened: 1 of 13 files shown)


crates/runestick/src/modules/tests.rs
```diff
@@ -0,0 +1,45 @@
+//! Utilities related to testing
```
udoprog, line 1: Since this adds a couple of dependencies and is currently only used for testing, could we move it to `rune-tests` instead?

crates/runestick/src/modules/char.rs
```diff
@@ -1,13 +1,25 @@
...
 /// Construct the `std::char` module.
 pub fn module() -> Result<Module, ContextError> {
     let mut module = Module::with_crate_item("std", &["char"]);
     module.ty::<ParseCharError>()?;
+    module.function(&["from_int"], char_from_int_impl)?;
     Ok(module)
 }
 
+fn char_from_int_impl(value: Value) -> Result<Option<Value>, VmError> {
+    let inner: i64 = value.into_integer()?;
```
udoprog, line 15: Rune can do this conversion if you use `i64` as an argument directly. Or `value: i64`.

crates/runestick/src/modules/tests.rs
tgolsson: Yes, sure!

crates/runestick/src/modules/char.rs
tgolsson: Ah, yes. I wanted to capture conversion errors but doesn't matter here.

udoprog:
Excellent, thank you!

PR #196, opened 2020-12-04
add generator::iter, clarify move/clone/reference for algebraic types

tgolsson:
This adds generator::iter and INTO_ITER protocols.

Also, I noticed that the clone vs auto-ref vs move behaviour of Options/Result was unclear and so I've opted to implement `as_ref()` using an autoboxing approach. This isn't... optimal, performance-wise, but should be safe.

As far as I can tell this should work already using the defined ToValue methods for Option<T: ToValue> and &Value, but that code path causes the SharedBox to drop/inc out of order and I'm not sure why - and this is more explicit. It's a bit lewd to write to self in what should be a pure operation but it works and is safe.

### DIFF
--- a/crates/runestick/src/from_value.rs
+++ b/crates/runestick/src/from_value.rs
@@ -141,6 +141,19 @@ impl UnsafeFromValue for &mut Option<Value> {
     }
 }
 
+impl UnsafeFromValue for &mut Result<Value, Value> {
+    type Output = *mut Result<Value, Value>;
+    type Guard = RawMut;
+
+    fn from_value(value: Value) -> Result<(Self::Output, Self::Guard), VmError> {
+        Ok(Mut::into_raw(value.into_result()?.into_mut()?))
+    }
+
+    unsafe fn unsafe_coerce(output: Self::Output) -> Self {
+        &mut *output
+    }
+}
+
 // String impls
 
 impl FromValue for String {
--- a/crates/runestick/src/modules/result.rs
+++ b/crates/runestick/src/modules/result.rs
@@ -42,6 +42,7 @@ fn and_then_impl(
     then: Function,
 ) -> Result<Result<Value, Value>, VmError> {
     match this {
+        // No need to clone v, passing the same reference forward
         Ok(v) => Ok(then.call::<_, _>((v,))?),
         Err(e) => Ok(Err(e.clone())),
     }
@@ -49,6 +50,7 @@ fn and_then_impl(
 
 fn map_impl(this: &Result<Value, Value>, then: Function) -> Result<Result<Value, Value>, VmError> {
     match this {
+        // No need to clone v, passing the same reference forward
         Ok(v) => Ok(Ok(then.call::<_, _>((v,))?)),
         Err(e) => Ok(Err(e.clone())),
     }
--- a/crates/runestick/src/vm.rs
+++ b/crates/runestick/src/vm.rs
@@ -2322,14 +2322,14 @@ impl Vm {
         let value = self.stack.pop()?;
 
         let value = match value {
-            Value::Option(option) => match option.take()? {
-                Some(value) => value,
+            Value::Option(option) => match &*(option.borrow_ref()?) {
+                Some(value) => value.clone(),
                 None => {
                     return Err(VmError::from(VmErrorKind::UnsupportedUnwrapNone));
                 }
             },
-            Value::Result(result) => match result.take()? {
-                Ok(value) => value,
+            Value::Result(result) => match &*(result.borrow_ref()?) {
+                Ok(value) => value.clone(),
                 Err(err) => {
                     return Err(VmError::from(VmErrorKind::UnsupportedUnwrapErr {
                         err: err.type_info()?,
--- a/crates/runestick/src/vm.rs
+++ b/crates/runestick/src/vm.rs
@@ -2322,13 +2322,13 @@ impl Vm {
         let value = self.stack.pop()?;
 
         let value = match value {
-            Value::Option(option) => match &*(option.borrow_ref()?) {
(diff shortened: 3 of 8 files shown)


crates/runestick/src/modules/result.rs
```diff
@@ -38,18 +40,33 @@ fn expect_impl(result: Result<Value, Value>, message: &str) -> Result<Value, VmE
...
-fn map_impl(this: &Result<Value, Value>, then: Function) -> Result<Result<Value, Value>, VmError> {
+fn map_impl(this: Result<Value, Value>, then: Function) -> Result<Result<Value, Value>, VmError> {
     match this {
         Ok(v) => Ok(Ok(then.call::<_, _>((v,))?)),
-        Err(e) => Ok(Err(e.clone())),
+        Err(e) => Ok(Err(e)),
     }
 }
+
+fn as_ref_impl(result: &mut Result<Value, Value>) -> Result<Result<Value, Value>, VmError> {
```
udoprog, line 59: I believe the effective outcome of this function is the same as cloning `result` and its variants twice.

1. `&result` auto-derefs to a reference of result and we pattern match over its variants.
2. The implementation of `ToValue` for `&Value` [is to clone it](https://github.com/rune-rs/rune/blob/main/crates/runestick/src/to_value.rs#L43).
3. We clone the value in the variant when building a new variant.
4. We then clone the result again at the bottom of the fn?

So I think this effectively has the same effect if we replace it with `Ok(result.clone())` minus a bit of extra cloning unless I'm missing something?

crates/runestick/src/from_value.rs
```diff
@@ -141,6 +141,19 @@ impl UnsafeFromValue for &mut Option<Value> {
...
+    type Guard = RawMut;
+
+    fn from_value(value: Value) -> Result<(Self::Output, Self::Guard), VmError> {
+        Ok(Mut::into_raw(value.into_result()?.into_mut()?))
+    }
+
+    unsafe fn unsafe_coerce(output: Self::Output) -> Self {
+        &mut *output
+    }
+}
```
udoprog, line 155: These impls I don't mind adding for completeness sake :+1:

crates/runestick/src/modules/result.rs
udoprog, line 59: I believe the effective outcome of this function is the same as cloning `result` and its variants twice.

1. `&result` auto-derefs to a reference of result and we pattern match over its variants.
2. The implementation of `ToValue` for `&Value` [is to clone it](https://github.com/rune-rs/rune/blob/main/crates/runestick/src/to_value.rs#L43).
3. We clone the value in the variant when building a new variant.
4. We then clone the result again at the bottom of the fn?

So I think this effectively has the same effect if we replace it with `Ok(result.clone())` minus a bit of extra cloning unless I'm missing something?

crates/runestick/src/modules/result.rs
```diff
@@ -38,18 +40,33 @@ fn expect_impl(result: Result<Value, Value>, message: &str) -> Result<Value, VmE
...
 ) -> Result<Result<Value, Value>, VmError> {
     match this {
         Ok(v) => Ok(then.call::<_, _>((v,))?),
-        Err(e) => Ok(Err(e.clone())),
+        Err(e) => Ok(Err(e)),
     }
 }
 
-fn map_impl(this: &Result<Value, Value>, then: Function) -> Result<Result<Value, Value>, VmError> {
+fn map_impl(this: Result<Value, Value>, then: Function) -> Result<Result<Value, Value>, VmError> {
```
udoprog, line 52: Just to clarify,this means that `.map` will now take ownership of the Result type it maps over.

udoprog (commented):
Thanks! I'm fine with the other changes :+1: but I'd like to know what the overall goal is w/ the added `Result::as_ref`?

Two things to take note:
* The closes thing to a reference in Rune is the clone of a value, which means a refcount increase for complex types.
* We don't support references of primitive types (all though this might be supported in the future by boxing them).

tgolsson:
Yeah, this is probably me not understanding how Value works, but I think this requires some work. The "bad case" I found during day 4 was code like this:

```rust
fn do_something(v) {
	false
}

struct Data {
	field
}

fn validate(v) {
	if !do_something(v.field?) {
		println!("field failed: {:?}", v.field);
	}
}

pub fn main() {
	let d = Data { field: Some("hello") };
	validate(d)
}
```

 It's *correct*, but also... not what I wanted. And I found no way to do this easily without moving from the value, or explicitly cloning it. Maybe all operations on Option and Result should then take references on the Rust side and not values? And there's other cases where I'm not sure what the correct semantics would be but where I can't express what I want because I can't qualify how I want it to behave, e.g., 

```rust
let v = Some(10);
if let Some(v) = v {
     v += 10;
}
dbg(v); // 10
```

crates/runestick/src/modules/result.rs
tgolsson: Yeah, this isn't actually doing what I expected. What I wanted and thought I got is Shared, this still doesn't work for primitives. Bah.

crates/runestick/src/vm.rs
```diff
@@ -2322,14 +2322,14 @@ impl Vm {
         let value = self.stack.pop()?;
 
         let value = match value {
-            Value::Option(option) => match option.take()? {
-                Some(value) => value,
+            Value::Option(option) => match &*(option.borrow_ref()?) {
```
udoprog, line 2325: Parens not needed! (I think)

udoprog (approved):

udoprog:
Thanks! Unwrap taking the value is such an obvious mistake in hindsight!

crates/runestick/src/vm.rs
tgolsson: Seems so, will remove!

PR #199, opened 2020-12-06
add set difference, intersection, union, fix sum error op

tgolsson:
This adds `HashSet::{difference, intersection, union}` methods. To make that doable it adds more blanket implementations for `TypeOf` so we can take `self` as `Ref<Self>` in receiver position. This required removing one static type declaration. This required removing one static type declaration for `Ref<Function> - I can't really tell the extent of the impact that has, but that specific incantation seems to only be used in one FromValue declaration.

Also fixes a minor copy-paste error in Iterator summation leading to very confusing error messages.

### DIFF
--- a/crates/runestick/src/modules/collections.rs
+++ b/crates/runestick/src/modules/collections.rs
@@ -174,6 +174,12 @@ where
             }
         }
     }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (_, upper) = self.this.size_hint();
+        (0, upper)
+    }
 }
 
 struct Difference<I>
@@ -197,6 +203,12 @@ where
                 return Some(elt);
             }
         }
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (_, upper) = self.this.size_hint();
+        (0, upper)
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/runestick/src/modules/collections.rs
```diff
@@ -107,6 +107,123 @@ impl HashSet {
...
+
+impl<I> std::iter::Iterator for Intersection<I>
+where
+    I: std::iter::Iterator<Item = Key>,
+{
+    type Item = Key;
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            // guaranteed to leave here unless the iterator is unbounded
+            let elt = self.this.next()?;
```
udoprog, line 170: Maybe clear `self.other` once iteration is completed to release the reference?

udoprog (approved):

udoprog:
Looks good! Minor nonblocking nit.

tgolsson:
@udoprog this should do if I understood you correctly.

udoprog:
Yeah, exactly! I think the loop can be cleaned up a bit:

```rust
        let other = self.other.take()?;

        loop {
            let item = self.this.next()?;

            if other.set.contains(&item) {
                self.other = Some(other);
                return Some(item);
            }
        }
```
I.e. if `other` is not set, the iterator has completed. If `this` completes, then `other` is not restored.

Issue #207, opened 2020-12-08
Unreferenced public imports are stripped


tgolsson:
e.g,

```rust
pub use other_file::do_something;

pub fn main() {
     do_something; // this line does nothing but if you remove it `do_something` won't be externally visible.
}
```

This requires one to create bogus references to all public functions; as in https://github.com/tgolsson/aoc-2020/blob/main/scripts/main.rn#L11.

I've noticed similar issues with module-private functions being marked as unused if the main function isn't called main, e.g.,
https://github.com/tgolsson/aoc-2020/blob/main/scripts/day1.rn#L3 `filter_inner` is marked as unused by the LSP.

tgolsson:
@udoprog I'm happy with after your improvements, do you feel more work is needed or shall we close this? :-)

udoprog:
Ah, yeah, this can be closed.

PR #218, opened 2020-12-12
#[test] attribute and related changes

tgolsson:
This does the following:

* Adds #[test] as an function attribute
* Adds a --test flag to rune-cli, allowing tests to be executed
* Adds support for callstacks in Sync code
* Changes how the error messages are printed to split cause, location, and detail into different parts instead of always using "internal vm error" as the highlighted error.

I don't think there are any major issues with the current impl so I'd like to merge this (after some cleanup) and open a bunch of issues for follow-up work. If you think any of these are blockers for merging (or something else) I'm open to discussion! I'm starting to shoehorn more and more things into this PR because the main feature is done, which makes it harder to work with and review.

These are good topics for follow-up work: 

* Add support for cfg-flags for conditional compilation
* Rework rune-cli to have `test` as a subcommand instead of a flag (and add `rune run` as the default command)
  * Add a `--short` mode to tests (`...F..N.E...F` style)
  * Add stdout capture by default for tests and add `--no-capture`
* Add more detail to DebugInst with instruction-specific info, e.g., lhs/rhs spans for ops, and other things.
* Expose test output/running to rune-lsp (maybe dependent/complicates/prevents #73)
* Fix tree-shaking to make test-discovery easier (tracked in #207)
* Convert some tests to Rune level tests instead of testing with Rust


crates/rune-cli/src/main.rs
```diff
@@ -174,6 +178,10 @@ async fn try_main() -> Result<ExitCode> {
         options.parse_option(opt)?;
     }
 
+    if args.test {
+        options.test(true);
+    }
```
tgolsson, line 183: options.test(args.test)

crates/rune-cli/src/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+                return Err(VmError::from(VmErrorKind::MissingFunction { hash: test.0 }).into());
+            }
+        };
+
+        vm.set_ip(offset);
+        match vm.async_complete().await {
+            Err(e) => {
+                // TODO: store output here
+                failures.insert(test.1.item.item.clone(), FailureReason::Crash(e));
+                writeln!(out, "crashed")?;
```
tgolsson, line 71: "failed"

crates/rune-cli/src/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+                    .expect("failed writing info");
+            }
+            FailureReason::ReturnedNone => continue,
+            FailureReason::ReturnedErr(e) => {
+                println!("----------------------------------------");
+                println!("Test: {}\n", item);
+                println!("Return value: {:?}\n", e);
+            }
+        }
+    }
```
tgolsson, line 116: I think moving all of this information into a test-case struct would make this clearer.

crates/rune-cli/src/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    println!("====");
+    println!(
+        "Ran {} tests with {} failures in {:.3} seconds",
+        tests.len(),
+        failure_count,
+        elapsed.as_secs_f64()
+    );
```
tgolsson, line 124: Should this go above or below the detailed output?

crates/rune/src/diagnostics.rs
```diff
@@ -216,13 +220,90 @@ impl EmitDiagnostics for VmError {
...
+                labels.push(
+                    Label::primary(source_id, span.range())
+                        .with_message("in this expression".to_string()),
+                );
+
+                (
+                    format!("type mismatch for operation `{}`", op),
+                    vec![
+                        format!("left hand side has type `{}`", lhs),
+                        format!("rhs hand side has type `{}`", rhs),
```
tgolsson, line 239: `right`, not rhs

crates/rune/src/diagnostics.rs
```diff
@@ -216,13 +220,90 @@ impl EmitDiagnostics for VmError {
         let source_id = debug_inst.source_id;
         let span = debug_inst.span;
 
-        labels.push(Label::primary(source_id, span.range()).with_message(error.to_string()));
+        let mut backtrace = vec![StackFrame { source_id, span }];
+        let (reason, notes) = match error {
+            VmErrorKind::Panic { reason } => {
+                labels.push(Label::primary(source_id, span.range()).with_message("panicked"));
+                ("panic in Runetime".to_owned(), vec![reason.to_string()])
```
tgolsson, line 227: `runtime`

crates/rune/src/diagnostics.rs
```diff
@@ -216,13 +220,90 @@ impl EmitDiagnostics for VmError {
...
+            use codespan_reporting::files::Files;
+
+            let line = files
+                .line_index(frame.source_id, frame.span.start.into_usize())
+                .unwrap();
+            let line = files.line_number(frame.source_id, line).unwrap() - 1;
+            let line_range = files.line_range(frame.source_id, line).expect("a range");
+            let source = files.get(frame.source_id)?;
+            let name = source.name();
+            let slice = &source.source()[line_range];
```
tgolsson, line 303: cleanup this with a helper function

tgolsson: Maybe this should take the span from the inst instead... This is a bit odd right now with closures, you can end up repeating the same multiple times.

crates/rune/src/indexing/index.rs
```diff
@@ -593,14 +593,34 @@ impl Index for ast::ItemFn {
...
         let is_toplevel = idx.items.is_empty();
+
+        let is_test = match attributes.try_parse::<attrs::Test>()? {
+            Some(_test) => {
+                // if !is_toplevel {
+                //     return Err(CompileError::msg(
+                //         span,
+                //         "#[test] is not supported on nested",
+                //     ));
+                // }
```
tgolsson, line 611: remove this commented code

crates/rune/src/indexing/index.rs
```diff
@@ -712,7 +740,7 @@ impl Index for ast::ItemFn {
             };
 
             idx.query.insert_meta(span, meta)?;
-        } else if is_toplevel && item.visibility.is_public() {
+        } else if is_toplevel && item.visibility.is_public() || is_test {
```
tgolsson, line 743: this isn't enough to make #[test] in mods auto-discovered, we need to be even more eager about them.

crates/runestick/src/compile_meta.rs
```diff
@@ -194,6 +197,9 @@ pub enum CompileMetaKind {
     ConstFn {
         /// Opaque identifier for the constant function.
         id: Id,
+
+        /// Whether this function has a test annotation
+        is_test: bool,
```
tgolsson, line 202: @udoprog I think you had opinions on this. See https://github.com/rust-lang/rust/issues/51999 and related threads for rust background on static contexts and asserts. Seems like a reduced feature set is supported, which allows implementing tests as const fn. Not sure whether it is required to interact with the test system - I guess by definition any assert in const code blocks the build, which precludes testing it. It'd just become weird if you *want* to test the code and you end failing tests during compilation...

crates/runestick/src/item.rs
```diff
@@ -346,17 +346,21 @@ impl Item {
...
                 write!(f, "{}::", p)?;
             }
 
-            write!(f, "{}", last)
+            write!(f, "{}", last)?;
+
+            formatter.pad(&f)
         } else {
-            write!(f, "{{root}}")
+            formatter.pad("{root}")
```
tgolsson, line 363: This is ugly, but I found no better way - this makes sure padding specifiers in format strings work as intended.

udoprog:
Ok, merging it in since it's fairly isolated and patching it up in follow ups. Thanks!

PR #219, opened 2020-12-12
add `Vec::clone`, `int::abs`, `char::to_digit`

tgolsson:


crates/runestick/src/modules/int.rs
```diff
@@ -12,6 +12,7 @@ pub fn module() -> Result<Module, ContextError> {
     module.function(&["parse"], parse)?;
     module.function(&["max"], i64::max)?;
     module.function(&["min"], i64::min)?;
+    module.function(&["abs"], i64::abs)?;
```
udoprog, line 15: Huh, I think this is already added as an associated fn, which should mean it's registered under this name already.

PR #229, opened 2020-12-16
add {HashMap, HashSet}::clone, {HashMap, HashSet, Vec}::remove

tgolsson:
Also sorted the inst fns so it's a bit easier to see if something is already implemented.


udoprog (approved):
Very nice!

PR #233, opened 2020-12-21
add Vec::sort_by and String::cmp

tgolsson:


udoprog (approved):
Looks good, but is there any way to instantiate an `Ordering` from within Rune?

PR #234, opened 2020-12-22
Fix broken examples link in readme

maxmcd:


udoprog:
Thank you!

This also needs to be changed in https://github.com/rune-rs/rune/blob/main/crates/rune/src/lib.rs#L145 as well. Would really appreciate it if you could fix that!

Normally we run the [readmes.rn](https://github.com/rune-rs/rune/blob/main/tools/readmes.rn) tool to keep them in sync (after you've installed [`cargo-readme`](https://docs.rs/cargo-readme/3.2.0/cargo_readme/)):

```shell
cargo run --bin rune -- tools/readmes.rn
```

This generates all readmes based on what's in their corresponding `lib.rs` file.

PR #243, opened 2021-01-06
upgrade nanorand to 0.5.2

tgolsson:
nanorand generates incorrect random numbers before 0.5.1, see [RUSTSEC-2020-0089](https://github.com/RustSec/advisory-db/blob/master/crates/nanorand/RUSTSEC-2020-0089.md). It also doesn't support generating signed numbers any more, so this PR adds a workaround for that.

### DIFF
--- a/crates/rune-modules/Cargo.toml
+++ b/crates/rune-modules/Cargo.toml
@@ -36,7 +36,7 @@ reqwest = { version = "0.10.9", optional = true, default-features = false, featu
 tokio = { version = "0.2.22", optional = true }
 serde_json = { version = "1.0.60", optional = true }
 toml = { version = "0.5.7", optional = true }
-nanorand = { version = "0.4.4", optional = true, features = ["getrandom"] }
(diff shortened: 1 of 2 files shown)


udoprog:
Thanks!

Might be worth to use rand at least by default. Even if we don't provide much of a stability guarantee at the moment, it's currently a more vetted alternative. And it's already included as a transitive dependency.

tgolsson:
@udoprog Where did you find rand in the dependency tree?
```
➜ cargo tree | grep rand
│   │   ├── nanorand v0.5.2
│   │   │   └── getrandom v0.2.0
```

udoprog:
> @udoprog Where did you find rand in the dependency tree?
> 
> ```
> ➜ cargo tree | grep rand
> │   │   ├── nanorand v0.5.2
> │   │   │   └── getrandom v0.2.0
> ```

Oh. I thought it was part of the `reqwest` dependency tree. I guess not!

PR #245, opened 2021-01-08
add lifetime to SourceLoader

tgolsson:
This let's you use a SourceLoader with associated lifetimes. The reason it originally didn't work to add a lifetime was that I did it for both the CompileVisitor and SourceLoader. I ended up just skipping the visitor for now which sidesteps the MacroContext problem; but it's not very nice to have the two Rc's have different constraints (though there is also much less need for the visitor to have a borrow, so... meh).


udoprog (approved):
Looks good to me. If you want to change into a reference, by all means! Or merge as-is.

PR #246, opened 2021-01-14
Fix miscompilation by realizing that struct literal construction potentially requires cleanup

### LINKED ISSUE #200: Using `?` can lead to stack corruption in non-linear code.
This is an interesting case:

```rust 
struct Foo {
	x,
	y,
}

struct Bar {
	x, 
	y,
}

impl Bar {
	fn from_foo(foo) {
		Some(Self {
			x: foo.x?,
			y: foo.y?,
		})
	}
}

pub fn main() {
	let f = Foo { x: Some(1), y: None };
	Bar::from_foo(f)
}
```

If the first element is Some, any following `None` will cause the VM to error out:

```
➜ cargo run --bin rune -- --dump-instructions -t miscomp.rn
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/rune --dump-instructions -t miscomp.rn`
# instructions
fn main() (0xde7d86e18013c5c8):
  0000 = push 1
  0001 = call 0xf39f074fddf64c6b, 1 // variant ::std::option::Option::Some
  0002 = call 0x68988acb6b2de244, 0 // variant ::std::option::Option::None
  0003 = struct 0x30d77e445d4bc057, 0
  0004 = copy 0 // var `f`
  0005 = call 0x94fbe6278aba2ca5, 1 // fn Bar::from_foo
  0006 = clean 1
  0007 = return

fn Bar::from_foo(foo) (0x94fbe6278aba2ca5):
  0008 = copy 0 // var `foo`
  0009 = object-index-get 0
  0010 = dup
  0011 = is-value
  0012 = jump-if 2 // label:try_not_error_2
  0013 = clean 1
  0014 = return
try_not_error_2:
  0015 = unwrap
  0016 = copy 0 // var `foo`
  0017 = object-index-get 1
  0018 = dup
  0019 = is-value
  0020 = jump-if 2 // label:try_not_error_3
  0021 = clean 1
  0022 = return
try_not_error_3:
  0023 = unwrap
  0024 = struct 0x9a4ef349214d9ab1, 0
  0025 = call 0xf39f074fddf64c6b, 1 // variant ::std::option::Option::Some
  0026 = clean 1
  0027 = return
# strings
0xcd01b01a74097893 = "x"
0xef5992decfa6104f = "y"
# object keys
0 = ["x", "y"]
fn main() (0xde7d86e18013c5c8):
  0000 = push 1
  0001 = call 0xf39f074fddf64c6b, 1 // variant ::std::option::Option::Some
  0002 = call 0x68988acb6b2de244, 0 // variant ::std::option::Option::None
  0003 = struct 0x30d77e445d4bc057, 0
  0004 = copy 0 // var `f`
  0005 = call 0x94fbe6278aba2ca5, 1 // fn Bar::from_foo
fn Bar::from_foo(foo) (0x94fbe6278aba2ca5):
  0008 = copy 0 // var `foo`
  0009 = object-index-get 0
  0010 = dup
  0011 = is-value
  0012 = jump-if 2 // label:try_not_error_2
try_not_error_2:
  0015 = unwrap
  0016 = copy 0 // var `foo`
  0017 = object-index-get 1
  0018 = dup
  0019 = is-value
  0020 = jump-if 2 // label:try_not_error_3
  0021 = clean 1
  0022 = return
== ! (stack error: tried to access out-of-bounds stack entry (at inst 22)) (231.437µs)
error: virtual machine error
   ┌─ miscomp.rn:15:20
   │
15 │                 y: foo.y?
   │                    ^^^^^^ stack error: tried to access out-of-bounds stack entry

```

As far as I can tell this comes from the fact that the original foo is still on the stack when returning:

```
[2020-12-06T20:05:05Z TRACE runestick::vm] 21: clean 1
    1+0 = Foo { x: Some(1), y: None }
    1+1 = None
  miscomp.rn:15  -                 y: foo.y?
  0022 = return
```

and it should likely be cleaned here as well too.

robojumper:
Fixes #200. I shortened the test case for brevity.

### DIFF
--- a/crates/rune/src/compiling/v1/assemble/expr_object.rs
+++ b/crates/rune/src/compiling/v1/assemble/expr_object.rs
@@ -5,6 +5,8 @@ use crate::compiling::v1::assemble::prelude::*;
 impl Assemble for ast::ExprObject {
     fn assemble(&self, c: &mut Compiler<'_>, needs: Needs) -> CompileResult<Asm> {
         let span = self.span();
+        let guard = c.scopes.push_child(span)?;
+
         log::trace!("ExprObject => {:?} {:?}", c.source.source(span), needs);
 
         let mut keys = Vec::<Box<str>>::new();
@@ -38,6 +40,7 @@ impl Assemble for ast::ExprObject {
                 let var = c.scopes.get_var(&*key, c.source_id, span)?;
                 var.copy(&mut c.asm, span, format!("name `{}`", key));
             }
+            c.scopes.decl_anon(span)?;
         }
 
         let slot = c.unit.new_static_object_keys_iter(span, &keys)?;
@@ -85,6 +88,7 @@ impl Assemble for ast::ExprObject {
             c.asm.push(Inst::Pop, span);
         }
 
+        c.scopes.pop(guard, span)?;
         Ok(Asm::top(span))
     }
 }
--- a/tests/tests/vm_try.rs
+++ b/tests/tests/vm_try.rs
@@ -20,6 +20,20 @@ fn test_unwrap() {
     };
 
     assert_eq! {
+        rune! { Option<bool> =>
+            struct Bar {
+                x, 
+                y,
+            }
+            
+            pub fn main() {
+                (Bar{ x: Some(1), y: None? }).x
+            }            
+        },
+        None,
+    };
+
+    assert_eq! {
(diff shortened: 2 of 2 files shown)


udoprog:
Brilliant! Thank you very much!

PR #250, opened 2021-03-13
Defined Result as std::result::Result to prevent bug

Roba1993:
To remove the bug of dervice Any is not working when a project defines a own Result implementation, this PR is calling out all Results as std::result::Result to prevent this bug.

### DIFF
--- a/crates/runestick-macros/src/context.rs
+++ b/crates/runestick-macros/src/context.rs
@@ -489,7 +489,7 @@ impl Context {
         name: &TokenStream,
         install_with: &TokenStream,
         tokens: &Tokens,
-    ) -> std::result::Result<TokenStream, Vec<syn::Error>>
+    ) -> Result<TokenStream, Vec<syn::Error>>
     where
         T: Copy + ToTokens,
     {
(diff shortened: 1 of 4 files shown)


crates/runestick-macros/src/context.rs
```diff
@@ -489,7 +489,7 @@ impl Context {
         name: &TokenStream,
         install_with: &TokenStream,
         tokens: &Tokens,
-    ) -> Result<TokenStream, Vec<syn::Error>>
+    ) -> std::result::Result<TokenStream, Vec<syn::Error>>
```
udoprog, line 492: So you don't have to change these ones, because they're part of the proc macro. We only need to change the ones which are part of a macro expansion. This usually just means anything that's in a call to `quote!` or `quote_in!`.

crates/runestick-macros/src/context.rs
```diff
@@ -521,7 +521,7 @@ impl Context {
             }
 
             impl #install_into_trait for #ident {
-                fn install_with(module: &mut #module) -> Result<(), #context_error> {
+                fn install_with(module: &mut #module) -> std::result::Result<(), #context_error> {
```
udoprog, line 524: Use `::std::result::Result`. It would guard against the case where someone defines their own local module called `std` as well!

udoprog (commented):
Thanks a bunch for finding the issue!

crates/runestick-macros/src/context.rs
Roba1993: Okay I change that

crates/runestick-macros/src/context.rs
Roba1993: Yep makes sense

PR #258, opened 2021-04-06
Implement OOTB template literal for bool and byte fixes #252

### LINKED ISSUE #252: OOTB template literal are missing for bool
There is no OOTB template literal implemented for the bool type.
This should be done in my opnion, because i64 etc. have an template literal.

Roba1993:
Implements the template literal for bool and byte.

Please have a view on it.

### DIFF
--- a/crates/runestick/src/value.rs
+++ b/crates/runestick/src/value.rs
@@ -8,6 +8,7 @@ use crate::{
 use serde::{de, ser, Deserialize, Serialize};
 use std::cmp;
 use std::fmt;
+use std::fmt::Write;
 use std::hash;
 use std::sync::Arc;
 use std::vec;
@@ -325,10 +326,10 @@ impl Value {
                 s.push_str(buffer.format(*float));
             }
             Value::Bool(bool) => {
-                s.push_str(&format!("{}", bool));
+                write!(s, "{}", bool)?;
             }
             Value::Byte(byte) => {
-                s.push_str(&format!("{:#04X}", byte));
+                write!(s, "{:#04X}", byte)?;
             }
             value => {
                 let b = Shared::new(std::mem::take(s));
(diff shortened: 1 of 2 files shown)


crates/runestick/src/value.rs
```diff
@@ -324,6 +324,12 @@ impl Value {
                 let mut buffer = ryu::Buffer::new();
                 s.push_str(buffer.format(*float));
             }
+            Value::Bool(bool) => {
+                s.push_str(&format!("{}", bool));
```
udoprog, line 328: Note that `format!` heap allocates a `String`. This can be avoided by doing something like this instead:
```rust
return Ok(write!(s, "{}", bool));
```
Same below w/ `Byte`.

udoprog:
Cheers for the patch! One comment that should be addressed.

crates/runestick/src/vm_error.rs
```diff
@@ -176,6 +176,11 @@ pub enum VmErrorKind {
     Halted { halt: VmHaltInfo },
     #[error("failed to format argument")]
     FormatError,
+    #[error("standard format error: {error}")]
+    StdFormatError {
+        #[from]
+        error: std::fmt::Error,
+    },
```
udoprog, line 183: Look closely at the signature of `string_display_with`: `Result<fmt::Result, VmError>`. That means that it expects the `Ok` variant to be a `fmt::Result`, which is exactly what `write!` returns. The idea here is that formatting should propagate an error (which can be caught by user code) instead of a `VmError` (which cannot).

Roba1993: Ah damn, stupid me....

udoprog: Not at all! It is a bit subtle.

udoprog:
Much obliged!

PR #267, opened 2021-05-28
Added file extension to introduction

TheLeonsver1:
Added the file extension to the introduction so it will be harder to miss.


udoprog:
Sorry for the late review.

I'm not entirely happy about the location of this. It sort of comes in the middle of nowhere. It would be better if this was added as a hint on the example of the first script file so that the reader has something to anchor it with, like here:
https://rune-rs.github.io/book/getting_started.html

udoprog (approved):

PR #268, opened 2021-06-06
add support for deriving Any on generic types

tgolsson:
I'm not super happy with the way `exact` is implemented, but nothing constant can be done when using generics. Apart from that, all this does is propagate the type constraints etc to all generated impls. The only major change is the `Named` trait which specializes the `exact` to generate the type name.

### DIFF
--- a/crates/runestick-macros/src/context.rs
+++ b/crates/runestick-macros/src/context.rs
@@ -17,6 +17,7 @@ struct Generate<'a> {
     field_ident: &'a syn::Ident,
     ty: &'a syn::Type,
     name: &'a syn::LitStr,
+    ty_generics: &'a syn::TypeGenerics<'a>,
 }
 
 pub(crate) struct FieldProtocol {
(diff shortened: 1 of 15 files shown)


crates/runestick-macros/src/context.rs
```diff
@@ -487,61 +492,84 @@ impl Context {
...
+        let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+
+        let generic_names = generics.type_params().map(|v| &v.ident).collect::<Vec<_>>();
+
+        let impl_named = if !generic_names.is_empty() {
+            quote! {
+                impl #impl_generics #named for #ident #ty_generics #where_clause {
+                    const NAME: #raw_str  = #raw_str::from_str(#name);
+
+                    fn exact() -> String {
```
udoprog, line 532: `fn full_name` perhaps? Conversely we can rename the constant `BASE_NAME`.

It would also be useful to support some form of formatting protocol for `full_name` (taking `buf: &mut String` would be the easiest), since otherwise each recursion into `exact` could end up allocating while it doesn't really have to.

udoprog (changes requested):
Looks good. Just minor changes from my side.

crates/runestick-macros/src/context.rs
tgolsson: Part 1: Sure!

The second one sounds like premature optimization for me, but I'll do it if you insist :-) As it is right now, it's only used in the Module::ty call, which is once per application lifetime and type. That's... Unlikely to be the performance bottleneck any time soon. ;-)

tgolsson:
@udoprog Updated from your feedback, also added the test module.

PR #269, opened 2021-06-11
Enable passing external guards into Rune

tgolsson:
The specific use-case for me is to box up a Mutex guard and drop it when the inner goes out of scope, but this is essentially just an "on-dropped" callback with user-data... Maybe the bar should be a bit higher to prevent abuse.  I think with some cleanup this'll be something like `Shared::from_{ref_mut}_with_guard`, but for now I've made a bunch of things public. Having so many core types private makes it really hard to hack at this kind of API (because it needs to be in the slow-to-compile Rune instead of a client crate.

### DIFF
--- a/crates/runestick/src/any_obj.rs
+++ b/crates/runestick/src/any_obj.rs
@@ -4,6 +4,7 @@ use crate::{Any, Hash, RawStr};
 use std::any;
 use std::fmt;
 use std::mem::ManuallyDrop;
+use std::ops::{Deref, DerefMut};
 use thiserror::Error;
 
 /// Errors raised during casting operations.
@@ -112,6 +113,49 @@ impl AnyObj {
         }
     }
 
+    /// Construct an Any that wraps a Deref type, behaving as the Target of
+    /// the Deref implementation
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that the returned `AnyObj` doesn't outlive the
+    /// dereference target.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use runestick::{Any, AnyObj};
+    /// use std::cell::RefCell;
+    ///
+    /// #[derive(Any)]
+    /// struct Foo(u32);
+    ///
+    /// let mut v = RefCell::new(Foo(1u32));
+    /// let mut guard = v.borrow();
+    ///
+    /// let any = unsafe { AnyObj::from_deref(guard) };
+    ///
+    /// let b = any.downcast_borrow_ref::<Foo>().unwrap();
+    /// assert_eq!(b.0, 1u32);
+    /// ```
+    pub unsafe fn from_deref<T, U: Deref<Target = T>>(data: U) -> Self
+    where
+        T: Any,
+    {
+        let boxed_guard = Box::into_raw(Box::new(data));
+        Self {
+            vtable: &AnyObjVtable {
+                kind: AnyObjKind::RefPtr,
+                drop: drop_impl::<U>,
+                as_ptr: as_ptr_deref_impl::<T, U>,
+                debug: debug_ref_impl::<T>,
+                type_name: type_name_impl::<T>,
+                type_hash: type_hash_impl::<T>,
+            },
+            data: boxed_guard as *const _ as *const (),
+        }
+    }
+
     /// Construct an Any that wraps a mutable pointer.
     ///
     /// # Safety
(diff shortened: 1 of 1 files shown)


tgolsson:
This is an example of how I'm doing it in the client crate at the moment:

```rust
impl Transform {
    fn get_mut(world: &WorldProxy, entity: &EntityProxy) -> Result<Option<Value>, VmError> {
        let comps = world.world.get_mut::<Components<Transform>>().expect("TODO[TSolberg]");

        let comp = AtomicRefMut::filter_map(comps, |comps| comps.get_mut(entity.entity));
        let comp = match comp {
            Some(e) => e,
            None => {
                return Ok(None);
            }
        };

        let (value, guard) = unsafe { comp.into_raw() };

        let ptr = Box::into_raw(Box::new(guard));

        let inner = NonNull::from(Box::leak(Box::new(SharedBox {
            access: Access::new(true),
            count: Cell::new(1),
            data: unsafe { AnyObj::from_mut(value).into() },
            guard: Some(RawDrop {
                data: ptr as *const (),
                drop_fn: |user_data| unsafe {
                    drop(Box::from_raw(user_data as *mut AtomicBorrowRefMut));
                },
            }),
        })));

        let shared = Shared { inner };
        Ok(Some(shared.into()))
    }
}
```

udoprog:
Neat. All though I'm not entirely happy about making so many raw edges public!

My immediate thought is that this would probably fit better in `AnyObj`, which already uses a customizable vtable that I think is flexible enough to implement what you need. Something with a public API like this:

```rust
let any_obj = AnyObj::from_pointer_and_guard(value, guard);
```

What it needs is that the data pointer points to a combination of the pointer and the owned guard:

```rust
struct PointerAndGuard {
    value: *const (),
    guard: *const (),
}
```

And an `as_ptr` implementation which checks for the type of `Value`, and then deferences the `PointerAndGuard` struct:

```rust
/// `T` here would be the type of the value being pointed to.
fn pointer_and_guard_as_ptr_impl<T>(this: *const (), expected: Hash) -> Option<*const ()>
where
    T: Any,
{
    if expected == Hash::from_type_id(any::TypeId::of::<T>()) {
        let this = this as *const PointerAndGuard<T>;
        Some((*this).value)
    } else {
        None
    }
}
```

Finally a drop implementation which just drops the guard. But I'll leave that as an exercise to the reader ;)

tgolsson:
Thanks for the feedback @udoprog! I revised and simplifed a lot. Much simpler and less dangerous. I'll take another pass on documentation tomorrow and then this should be G2G.

crates/runestick/src/any_obj.rs
```diff
@@ -168,7 +208,54 @@ impl AnyObj {
...
+    /// unsafe {
+    ///     let shared = Shared::new(AnyObj::from_mut_deref::<Thing, _>(guard));
+    ///     shared.downcast_borrow_mut::<Thing>().unwrap().0 = 20;
+    ///     assert_eq!(20u32, shared.downcast_borrow_mut::<Thing>().unwrap().0);
+    ///     assert_eq!(20u32, shared.downcast_borrow_ref::<Thing>().unwrap().0);
+    ///     drop(shared);
+    ///     value.borrow();
+    /// }
+    /// ```
+    pub unsafe fn from_mut_deref<T, U: std::ops::DerefMut<Target = T>>(data: U) -> Self
```
udoprog, line 243: No harm in importing `DerefMut` into scope.
Also `from_deref_mut` to follow suggestion as above.

crates/runestick/src/any_obj.rs
```diff
@@ -1,9 +1,9 @@
 //! Helper types for a holder of data.
 
 use crate::{Any, Hash, RawStr};
-use std::any;
 use std::fmt;
 use std::mem::ManuallyDrop;
+use std::{any, ops::Deref};
```
udoprog, line 6: Separate line.

crates/runestick/src/any_obj.rs
```diff
@@ -168,7 +208,54 @@ impl AnyObj {
...
+        }
+    }
+
+    /// Construct an Any that wraps a mutable pointer. This overload lets you
+    /// pass in a lock or other guard-type, as well as the raw value.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that the returned `AnyObj` doesn't outlive the
+    /// reference it is wrapping and the source of the look.
```
udoprog, line 221: "source of the look"?

crates/runestick/src/any_obj.rs
```diff
@@ -112,6 +112,46 @@ impl AnyObj {
...
+    /// #[derive(Any)]
+    /// struct Foo(u32);
+    ///
+    /// let mut v = RefCell::new(Foo(1u32));
+    /// let mut guard = v.borrow();
+    /// let any = unsafe { AnyObj::from_ref_deref::<Foo, _>(guard) };
+    /// let b = any.downcast_borrow_ref::<Foo>().unwrap();
+    /// assert_eq!(b.0, 1u32);
+    /// ```
+    pub unsafe fn from_ref_deref<T, U: Deref<Target = T>>(data: U) -> Self
```
udoprog, line 137: Just `from_deref` is sufficient.

udoprog (approved):
Really neat to piggyback on `Deref` / `DerefMut`. Gives the function a really broad set of utility.

I need to take an even harder think about any potential bad things that can happen due to leaking a lifetimed guard like this. But since the function *is* `unsafe` and is primarily intended to be used in limited scope-like manners we're *probably* good to go for now.

A few style and naming nits. Other than that, merge when you feel like it. Awesome work!

udoprog:
Pushed one more change which removed `U` as a type parameters. I realized it's not needed since `Target` is an associated type. If you're cool with it, feel free to merge. :+1:

PR #271, opened 2021-07-09
Did all new Clippy Required Fixes

genusistimelord:
I have gone thru and updated everything to follow the latest Clippy rules to the best of my ability. Hopefully some of the changes can help boost speed due to removing double conversions or a Ref of a object that gets instantly de-referenced.


schungx:
>  Hopefully some of the changes can help boost speed due to removing double conversions or a Ref of a object that gets instantly de-referenced.

I believe if it is found by clippy, it would have been compiled out and optimized anyway.

But I find that it is always a best practice to eliminate _all_ clippy errors.  This way, you find that 1% of the cases where there was a hidden logic bug while 99% of the cases are usually annoyances.

PR #276, opened 2021-08-28
Fixed Clippy Issues and Added Allows for those we will Ignore

genusistimelord:
I have gone through and fixed all the Clippy issues within each crate and separated each Clippy issue fix into its own Commits.


udoprog:
Perfect. Thanks for doing the work and excluding the lints with undesirable effects!

PR #278, opened 2021-09-02
add support for "lightweight" function pointers

tgolsson:
I've had a bunch of use-cases where I want to store callbacks etc to Rune functions inside Rust, and the regular Function objects has a lot of baggage and works badly in threaded contexts (being neither Sync nor Send iirc). These raw functions behave like `fn` in Rust, as opposed to Fn, making them easier to store. I'm a bit annoyed at the code duplication and considered making these a sub=enum of the regular Function hierarchy, but I think this implementation is *cleaner*.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,21 @@ jobs:
           components: rustfmt
       - run: cargo fmt --all -- --check
 
+  build_documentation:
+    name: Build Documentation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+          default: true
+      - run: cargo doc --package rune
+        env:
+          RUSTFLAGS: --cfg docsrs
+          RUSTDOCFLAGS: --cfg docsrs
+
   build_wasm:
     name: Build Wasm
     runs-on: ubuntu-latest
--- a/crates/rune/src/runtime/function.rs
+++ b/crates/rune/src/runtime/function.rs
@@ -10,14 +10,351 @@ use std::future::Future;
 use std::sync::Arc;
 
 /// A callable non-sync function.
-pub type Function = FunctionImpl<Value>;
+#[repr(transparent)]
+pub struct Function(FunctionImpl<Value>);
+
+impl Function {
+    /// Perform an asynchronous call over the function which also implements
+    /// [Send].
+    pub async fn async_send_call<A, T>(&self, args: A) -> Result<T, VmError>
+    where
+        A: Send + Args,
+        T: Send + FromValue,
+    {
+        self.0.async_send_call(args).await
+    }
+
+    /// Perform a call over the function represented by this function pointer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rune::{Hash, Vm, FromValue};
+    /// use rune::runtime::Function;
+    /// use std::sync::Arc;
+    ///
+    /// # fn main() -> rune::Result<()> {
+    /// let mut sources = rune::sources! {
+    ///     entry => {
+    ///         fn add(a, b) {
+    ///             a + b
+    ///         }
+    ///
+    ///         pub fn main() { add }
+    ///     }
+    /// };
+    ///
+    /// let unit = rune::prepare(&mut sources).build()?;
+    /// let mut vm = Vm::without_runtime(Arc::new(unit));
+    /// let value = vm.call(&["main"], ())?;
+    ///
+    /// let value = Function::from_value(value)?;
+    /// assert_eq!(value.call::<_, u32>((1, 2))?, 3);
+    /// # Ok(()) }
+    /// ```
+    pub fn call<A, T>(&self, args: A) -> Result<T, VmError>
+    where
+        A: Args,
+        T: FromValue,
+    {
+        self.0.call(args)
+    }
+
+    /// Call with the given virtual machine. This allows for certain
+    /// optimizations, like avoiding the allocation of a new vm state in case
+    /// the call is internal.
+    ///
+    /// A stop reason will be returned in case the function call results in
+    /// a need to suspend the execution.
+    pub(crate) fn call_with_vm(&self, vm: &mut Vm, args: usize) -> Result<Option<VmHalt>, VmError> {
+        self.0.call_with_vm(vm, args)
+    }
+
+    /// Create a function pointer from a handler.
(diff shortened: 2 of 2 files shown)


crates/runestick/src/function.rs
```diff
@@ -333,6 +355,51 @@ impl fmt::Debug for Function {
     }
 }
 
+impl RawImpl {
+    /// Perform a call over the function represented by this function pointer.
+    pub fn call<A, T>(&self, args: A) -> Result<T, VmError>
+    where
+        A: Args,
```
udoprog, line 362: Since it's sync, I believe this can use `GuardedArgs` if you want to support references: https://docs.rs/runestick/latest/runestick/trait.GuardedArgs.html

crates/runestick/src/function.rs
```diff
@@ -25,6 +25,15 @@ where
     inner: Inner<V>,
 }
 
+/// A callable non-sync raw.
+pub type RawFunction = RawImpl;
```
udoprog, line 29: Name `RawImpl` to `RawFunction` directly instead of a type alias? `RawImpl` would leak out due to this alias anyway but the doc might end up being a bit awkward.

udoprog (commented):
Two inline notes.

Something other than `into_raw` might be more appropriate. Note that we also have `SyncFunction` which are (almost) fully capable function pointers supporting `Send + Sync`: https://docs.rs/runestick/latest/runestick/type.SyncFunction.html

The `FnOffset` variant is to my understanding one of the larger variants in the function pointer implementation, so how much smaller does the proposed `RawFunction` really end up being here?

udoprog:
I also see that `SyncFunction` documentation isn't being properly generated due to it being a type alias :/. Opening an issue for that.

crates/runestick/src/function.rs
tgolsson: That sounds reasonable! This uses the same interface as the regular Function. Does it make sense to change the default interface as well?

crates/runestick/src/function.rs
tgolsson: Yeah, will do. Just followed the default pattern but when the parametrization of Value/ConstValue went away I was a bit torn on whether keeping the pattern was worthwhile.

tgolsson:
@udoprog Should I fix the docs when I'm at it?

udoprog:
> @udoprog Should I fix the docs when I'm at it?

Yes please!

tgolsson:
Re: SyncFunction, that got quite far but also wasn't Clone, and iirc solving that popped up a bunch of other problems.

tgolsson:
@udoprog Do you have any name suggestions that's not `raw`? :-) I guess closest is that it's a Ptr (or ptr-like) but that only makes sense as a suffix...

udoprog:
@tgolsson not sure! What is it about SyncFunction that makes it hard to implement `Clone`?

tgolsson:
> @tgolsson not sure! What is it about SyncFunction that makes it hard to implement `Clone`?

It wasn't the clone per se, but I still had some sync/send issues in my downstream crate after the original clone change. I've sat on the raw variant a few days and trying Clone+SyncFunction again now works -- Probably contextual in how I was using/referencing it causing problems. I've removed the "raw" part, so this is just the docs fix + Clone.

crates/runestick/src/function.rs
```diff
@@ -25,6 +26,139 @@ where
...
+    pub fn async_send_call<'a, A, T>(
+        &'a self,
+        args: A,
+    ) -> impl Future<Output = Result<T, VmError>> + Send + 'a
+    where
+        A: 'a + Send + Args,
+        T: 'a + Send + FromValue,
+    {
+        self.0.async_send_call(args)
+    }
```
udoprog, line 41: Hm. Isn't `Send` and `Sync` auto-implemented for the future if you do this?
```rust
    pub async fn async_send_call<A, T>(&self, args: A) -> Result<T, VmError>
    where
        A: 'a + Send + Args,
        T: 'a + Send + FromValue,
    {
        self.0.async_send_call(args).await
    }
```

tgolsson: Yes; but that's a breaking API change -- current one shouldn't be. I think.

udoprog: Oh. Hm. I'm fairly confident that changing to a newtype already is a break. And I wouldn't worry too much about it at this stage. We just release a new breaking release on the next release.

I'm also not sure this strictly is breaking. I think the future generated is compatible with the signature you proposed.

tgolsson: Yeah, true. Updated!

crates/runestick/src/function.rs
```diff
@@ -29,15 +29,12 @@ where
 impl Function {
     /// Perform an asynchronous call over the function which also implements
     /// `Send`.
-    pub fn async_send_call<'a, A, T>(
-        &'a self,
-        args: A,
-    ) -> impl Future<Output = Result<T, VmError>> + Send + 'a
+    pub async fn async_send_call<'a, A, T>(&'a self, args: A) -> Result<T, VmError>
```
udoprog, line 32: I think the lifetime can be elided here?

udoprog (approved):
One nit. This looks good to me now. I would like to see some tests as it's public API but I can tolerate adding them later if you don't want to do it now. Cheers!

tgolsson:
@udoprog I can do some light doctests, no problems! I think the only variant that can be somewhat easily constructed within runestick is "FnHandler" though, but I guess that's fine. I'll see if I can whip that up tomorrow!

udoprog:
> @udoprog I can do some light doctests, no problems! I think the only variant that can be somewhat easily constructed within runestick is "FnHandler" though, but I guess that's fine. I'll see if I can whip that up tomorrow!

I appreciate it! It might be simpler for a number of them to compile small rune programs (and I was sure there was a helper to do that somewhere?). Also, this PR can be merged at your own leisure with or without doctets.

tgolsson:
> I appreciate it! It might be simpler for a number of them to compile small rune programs (and I was sure there was a helper to do that somewhere?). Also, this PR can be merged at your own leisure with or without doctets.

I believe that helper is in the testing lib, which isn't a dependency of `runestick`? I think it used to reside in Rune pre some refactor.

udoprog:
> Edit: I still need to go that route or expose more of the API... It's not possible to create Functions manually as an external user since the Handler type and Function constructors are all crate-private.

The constructors are private, but they all should and could implement `FromValue` I think?

udoprog:
@tgolsson rebased and added test cases on the new branch. Please give it a 👍 if it works correctly for you!

PR #280, opened 2021-09-05
add `unwrap_or` impl to result.

tgolsson:


crates/runestick/src/modules/result.rs
```diff
@@ -11,6 +11,7 @@ pub fn module() -> Result<Module, ContextError> {
     module.inst_fn("is_ok", is_ok)?;
     module.inst_fn("is_err", is_err)?;
     module.inst_fn("unwrap", unwrap_impl)?;
+    module.inst_fn("unwrap_or", unwrap_or_impl)?;
```
udoprog, line 14: Hm. I think `Result::<Value, Value>>::unwrap_or` can be used as an alternative to `unwrap_or_impl`.

udoprog (approved):
LGTM. Cheers!

PR #287, opened 2021-11-03
Allow re-using a Vm by supporting mutable VmExecution

### LINKED ISSUE #286: How to call a function more than once?
I'd like to be able to call a function in Rune from Rust multiple times, as cheaply as possible.
I noticed `Vm::execute` and `Vm::call` take `self` by move, so it is not possible to reuse a vm after the call completed.
What is the idiomatic way to run a short function multiple times, with minimal FFI overhead?

udoprog:
Been dragging my feet on this since I mostly use Rune while calling a single function. But this adds the ability to perform multiple calls on the same virtual machine assuming you have mutable access to it.

The solution adds a generic parameter to `VmExecution` to hold the head virtual machine, which is constrained to implement `AsRef<Vm>` and `AsMut<Vm>` as appropriate. This allows `VmExecution` to both take an exlusive reference `VmExecution<&mut Vm>` and ownership (default) `VmExecution<Vm>`. All the corresponding functions on `Vm` that does take `self` by ownership has been changed to take `&mut self` instead where possible.

As far as I can tell, nothing bad should be possible - since mutable access implies exclusive access to the same degree as ownership does (minus that the code is not responsible for dropping the virtual machine). The only point I'm a little bit concerned about is `GuardedArgs` since it allows for passing in references during async function calls. But since everything must be pinned and due to the pinning guarantee there should be no way for the guard that is created to leak anyways.

Fixes #286 since the solution then becomes obvious. Re-use the same `Vm`.

### DIFF
--- a/benches/benches/aoc_2020_11a.rs
+++ b/benches/benches/aoc_2020_11a.rs
@@ -18,7 +18,7 @@ fn aoc_2020_11a(b: &mut Bencher) -> runestick::Result<()> {
         .map(str::to_owned)
         .collect::<Vec<String>>();
 
-    let vm = rune_tests::rune_vm! {
+    let mut vm = rune_tests::rune_vm! {
         enum CellState {
             Floor,
             Unoccupied,
@@ -244,7 +244,7 @@ fn aoc_2020_11a(b: &mut Bencher) -> runestick::Result<()> {
     let entry = runestick::Hash::type_hash(&["main"]);
 
     b.iter(|| {
-        let execution = vm.clone().execute(entry, (input.clone(),));
+        let execution = vm.execute(entry, (input.clone(),));
         let mut execution = execution.expect("successful setup");
         execution.complete().expect("successful execution")
     });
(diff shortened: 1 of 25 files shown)


udoprog:
So there was a handful of more things that needed to be taken care of to allow for `Vm` reuse.

First is 1c53a9d which ensures that the stack is always cleared by `VmExecution`. This ensures that the contract for `GuardedArgs` holds so that we can pass in references.

Second is 085672e which clears the stack of call frames. Since a Vm can be passed any kind of instructions, it might also be passed instructions which do not take care to balance call frames, or we might encounter a vm error which is propagated out of the virtual machine while preserving call traces to perform a backtraces. If such a `Vm` is re-used it could result in garbled execution, so we preemptively clear all call frames instead.

tgolsson (approved):
LGTM! It's a bit shame to have a downstream breakage because dependent crates will need to change every VM use to be mut... I guess the alternative is overloading with more functions, which seems confusing as well.

udoprog:
> LGTM! It's a bit shame to have a downstream breakage because dependent crates will need to change every VM use to be mut... I guess the alternative is overloading with more functions, which seems confusing as well.

Yeah. It's an interesting example of where the `mut` keyword in itself causes breakage in code that should work (if you can take ownership you can definitely get a mutable reference). It's a shame, but I don't think pre-1.0 projects should be too worried about API breakages that makes sense.

PR #291, opened 2021-11-05
Introduce Vm::with to allow for calling functions inside of the vm

### LINKED ISSUE #290: How to convert the return value into a STRING_DISPLAY string?
I'm calling a Rune function from Rust. That function can return an `Err(CassError)`, where `CassError` is an external type that implements the `STRING_DISPLAY` protocol.

Dumping that error object from inside of the script works fine:

```rust
pub async fn init(db) {
    let prep_result = db.prepare("insert", "a broken query").await;
    if let Err(e) = prep_result {
        dbg(`Error ${e}`);
    }
    prep_result
}
```

Then I try to print the `Value` I obtained from the call, and indeed it is a proper `Err` but after unwrapping I get only `Any(CassError)` description instead of the fully formatted object. 

```
"Error Cassandra detailed error SERVER_INVALID_QUERY: Invalid INTEGER constant (1) for \"id\" of type text"
error: Failed to initialize: Function init returned error: Any("CassError")
```

How do I get the formatted object outside of the vm? I tried the `string_display` method naively, hoping the docs is wrong, but it was right and it doesn't work outside of the vm :(

The source of `CassError` if that helps:
```rust
#[derive(Any, Debug)]
pub struct CassError {
    #[rune(get)]
    msg: String,
}

impl CassError {
    fn display(&self, buf: &mut String) -> std::fmt::Result {
        use std::fmt::Write;
        write!(buf, "{}", self.msg)
    }
}

impl Display for CassError {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.msg)
    }
}

impl std::error::Error for CassError {}

```

udoprog:
Introduces `Vm::with` as a public API for specifying which virtual machine to call a function inside. This maintains the necessary thread-locals to call the `Value` helper functions introduced earlier so that they can be publicly used.

Fixes #290

### DIFF
--- a/crates/runestick/src/value.rs
+++ b/crates/runestick/src/value.rs
@@ -286,11 +286,16 @@ pub enum Value {
 impl Value {
     /// Format the value using the [Protocol::STRING_DISPLAY] protocol.
     ///
-    /// Also requires a work buffer `buf` which will be used in case the value
-    /// provided required formatting.
+    /// Requires a work buffer `buf` which will be used in case the value
+    /// provided requires out-of-line formatting. This must be cleared between
+    /// calls and can be re-used.
     ///
-    /// Note that this function will always failed if called outside of a
-    /// virtual machine.
+    /// You must use [Vm::with] to specify which virtual machine this function
+    /// is called inside.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if called outside of a virtual machine.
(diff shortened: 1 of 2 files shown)


pkolaczk:
I confirm it works! Thank you so much!

udoprog:
Thanks for confirming!

PR #295, opened 2021-11-08
Don't require macro function to be Copy

pkolaczk:
Requiring the macro closure to be Copy disallows capturing things like `HashMap` or other complex structures which are `Clone` but not `Copy`. Changing the requirement to `Clone` still allows Rune to copy the closure if needed. 

Example where it could be useful:

```rust
/// Returns the literal value stored in the `params` map under the key given as the first 
/// macro arg, and if not found, returns the expression from the second arg.
fn param(params: &HashMap<String, String>, ts: &TokenStream) -> runestick::Result<TokenStream> {
...
}

...
let mut params = HashMap::new(); 
params.add("foo".to_string(), "1".to_string());

let mut my_module = Module::default();
my_module.macro_(&["param"], move |ts| Self::param(&params, ts)).unwrap();
...
```

Then I can use it in Rune script:

```rust
const FOO = param!("foo", -1);
```

### DIFF
--- a/crates/runestick/src/module.rs
+++ b/crates/runestick/src/module.rs
@@ -484,7 +484,7 @@ impl Module {
     /// Register a native macro handler.
     pub fn macro_<N, M, A, O>(&mut self, name: N, f: M) -> Result<(), ContextError>
     where
-        M: 'static + Send + Sync + Copy + Fn(&A) -> Result<O, crate::Error>,
+        M: 'static + Send + Sync + Clone + Fn(&A) -> Result<O, crate::Error>,
         A: any::Any,
         O: any::Any,
         N: IntoIterator,


udoprog:
Thank you! Since `Clone` does not appear to be needed I'd be OK with dropping it as well. All though that can of course wait.

pkolaczk:
Well, I left the `Clone` in case you ever want to make a copy of the closure later. This way you won't break the public API. If you removed `Clone` now and then people start relying on it not being `Clone`, adding it back later could break user's code.

udoprog:
Yeah. All though it's pretty safe to say we won't need it without a fundamental redesign of internals. All external callbacks are wrapped in `Arc` to make sure they can be cloned when needed.

PR #297, opened 2021-11-08
Make Try and Return more complex and remove Unwrap and IsValue

udoprog:
This change introduces the `Try` instruction which replaces the use of about 5 other instructions whenever an expression like `expr?` is used.

`Return` has also been extended to be able to process an addressed variable `address` and to `clean` (where `clean` indicates the number of stack elements that should be clean for the current call frame). This allows it to execute more efficiently under some circumstances.

So what happens is that for every try expression, we replace code that is generated like this:

```
  0000 = dup
  0001 = is-value
  0002 = jump-if 1 // label:try_not_error_4
  0003 = return
try_not_error_4:
  0004 = unwrap
```

With this:

```
  0000 = try address=top, clean=0, preserve=true
```

So for every try we save about 4 instructions. This example code goes from being 82 instructions to only be 14:

```
Ok(Ok(Ok(Ok(Ok(Ok(42))))))??????
```

We also avoid the work involved in advancing the Vm across 5 operations and can avoid performing some duplicate work, netting us a performance uplift of about 15% for try expressions.

Another benefit is that this allows us to nuke the `IsValue` and `Unwrap` instructions which were only used in try expressions.

### DIFF
--- a/crates/runestick/src/stack.rs
+++ b/crates/runestick/src/stack.rs
@@ -145,6 +145,15 @@ impl Stack {
         })
     }
 
+    /// Peek the given address without removing it from the stack if it's on the
+    /// top.
+    pub fn address_peek(&mut self, address: InstAddress) -> Result<&Value, StackError> {
+        match address {
+            InstAddress::Top => self.last(),
+            InstAddress::Offset(offset) => self.at_offset(offset),
+        }
+    }
+
     /// Pop the given number of elements from the stack.
     pub fn popn(&mut self, count: usize) -> Result<(), StackError> {
         drop(self.drain_stack_top(count)?);
--- a/crates/runestick/src/stack.rs
+++ b/crates/runestick/src/stack.rs
@@ -145,15 +145,6 @@ impl Stack {
         })
     }
 
-    /// Peek the given address without removing it from the stack if it's on the
-    /// top.
-    pub fn address_peek(&mut self, address: InstAddress) -> Result<&Value, StackError> {
-        match address {
-            InstAddress::Top => self.last(),
-            InstAddress::Offset(offset) => self.at_offset(offset),
-        }
-    }
-
     /// Pop the given number of elements from the stack.
     pub fn popn(&mut self, count: usize) -> Result<(), StackError> {
         drop(self.drain_stack_top(count)?);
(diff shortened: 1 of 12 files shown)


crates/runestick/src/vm.rs
```diff
@@ -2258,60 +2263,48 @@ impl Vm {
...
+            Value::Result(result) => match &*result.borrow_ref()? {
+                Result::Ok(value) => Some(value.clone()),
+                Result::Err(..) => None,
+            },
+            Value::Option(option) => match &*option.borrow_ref()? {
+                Option::Some(value) => Some(value.clone()),
+                Option::None => None,
+            },
             other => {
                 return Err(VmError::from(VmErrorKind::UnsupportedIsValueOperand {
```
tgolsson, line 2293: I think changing this error message would be prudent too?

crates/runestick/src/vm.rs
```diff
@@ -2258,60 +2263,48 @@ impl Vm {
...
         };
 
-        self.stack.push(is_value);
-        Ok(())
+        if let Some(value) = value {
+            if preserve {
+                self.stack.push(value);
+            }
+
+            Ok(false)
```
tgolsson, line 2304: I'd expect that we also popped the peeked address somewhere here? Or is it because we might unwrap something that isn't top that we can't?

crates/runestick/src/stack.rs
```diff
@@ -145,6 +145,15 @@ impl Stack {
         })
     }
 
+    /// Peek the given address without removing it from the stack if it's on the
+    /// top.
+    pub fn address_peek(&mut self, address: InstAddress) -> Result<&Value, StackError> {
```
tgolsson, line 150: A very minor thing but I try (and often forget) to share prefix of things that are similar instead of suffix. So peek/peek_address, pop/pop_address... I find it helps with code-completion and searching.

tgolsson (commented):
Very nice cleanup! Some minor things/nitpicks, apart from that looks good!

crates/runestick/src/vm.rs
udoprog: Dangit. I think you're right! If `InstAddress::Top` then it should be popped according to its contract.

udoprog: Fixed now. Thank you!

crates/runestick/src/vm.rs
udoprog: Two dedicated error messages introduced instead of both using the outdated one for IsValue. Thanks!

PR #303, opened 2021-11-11
Support registering new sources in macros and parsing in macros

### LINKED ISSUE #302: Support parsing string sources in macros
Currently there is no way to parse rune code from a string inside of a macro to produce a legal `TokenStream`. If the caller tried to use for example [`parse_all`](https://docs.rs/rune/0.9.1/rune/fn.parse_all.html) - once AST has been expanded it wouldn't reference the correct code but rather incorrectly assume that the spans being referenced belong to the source where the macro is invoked. Instead it should correctly be referencing a newly registered source. This is also a massive footgun we might want to look over!

To support this use case we'd have to enable macros to modify and add to the `Sources` collection (important for diagnostics) and for AST generated by the macro to correctly reference this new source. It might even be appropriate to start tagging tokens with source ids (in particular if they're absent) to ensure that expanded tokens do not accidentally reference the wrong source. In particular the `Resolve` trait should receive the full collection of `Sources` rather than just the *implicit* `Source` that corresponds to the source being processed.

See `Resolve`: https://github.com/rune-rs/rune/blob/331ba71f8faeb502f1c58d31c9696ffa4438abdb/crates/rune/src/parsing/resolve.rs#L83

In action here:
https://github.com/rune-rs/rune/blob/331ba71f8faeb502f1c58d31c9696ffa4438abdb/crates/rune/src/ast/lit_str.rs#L30

Relates to #300

udoprog:
Almost all of the changes presented in here follow from these requirements:
* All the necessary tokens that need to refer to a piece of source includes a `SourceId` and the lexer has been modified to include it where appropriate.
* The `Resolve` macro now takes the full collection of `Sources` instead of a single *implicit* `Source`.
* A mutable `Sources` is propagated into the macro context so that it can be modified during macro expansion.

Due to all of these, we can now implement `macros::parse_all` allowing to parse a string into a correctly behaving `TokenStream`.

Fixed #302

### DIFF
--- a/crates/rune/src/macros/functions.rs
+++ b/crates/rune/src/macros/functions.rs
@@ -1,7 +1,8 @@
 use crate::ir::{IrCompile, IrError, IrEval};
-use crate::macros::{current_context, ToTokens, TokenStream};
-use crate::parsing::{ResolveError, ResolveOwned};
+use crate::macros::{current_context, current_context_mut, ToTokens, TokenStream};
+use crate::parsing::{Parse, ParseError, ResolveError, ResolveOwned};
 use crate::Spanned;
+use runestick::Source;
 
 /// Evaluate the given target as a constant expression.
 ///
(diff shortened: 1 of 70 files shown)


crates/rune/src/macros/functions.rs
```diff
@@ -70,3 +71,20 @@ where
...
+pub fn parse_all<T>(source: &str) -> Result<T, ParseError>
+where
+    T: Parse,
+{
+    current_context_mut(|ctx| {
+        let sources = ctx.sources_mut();
+        let source_id = sources.insert(Source::new("macro", source));
+        crate::parse_all(source, source_id)
+    })
+}
```
udoprog, line 90: This is the `parse_all` implementation that can now be implemented inside of macros as a result of this change.

PR #313, opened 2021-11-14
replace manual de/ser with derive

tgolsson:
So; this proved to be very challenging -- Essentially, removing and/or changing ConstValue::StaticString impacts every API where ConstValue is involved in a ToValue/FromValue call, since one would have to pass a Unit around to resolve the proper index. I don't think it's worthwhile to rewrite all of this, as I'd argue a better approach is to (eventually) create a dedicated on-disk format which shields the runtime from these design decisions. I also don't understand the relationship between ConstValue/IrValue/Value enough to judge what is correct behaviour when these types are converted between each other, and which ones are valid.

My original interpretation is ConstValue was akin to literal (e.g., hardcoded value), and in the case of static string we shunt these to a `.data` segment, like a weak interning. However, this makes it wrong to convert from Value to ConstValue - but removing that breaks a lot of other code, so I think my assumptions are wrong.

Thus, I'll create two issues if this is merged:

* Document the interactions between Value/ConstValue and delimit their use patterns
* Create a dedicated unit format for serialization


crates/rune/src/runtime/bytes.rs
```diff
@@ -6,12 +6,13 @@ use crate::compile::{InstallWith, Named};
...
     FromValue, Mut, RawMut, RawRef, RawStr, Ref, UnsafeFromValue, Value, VmError,
 };
+use serde::{Deserialize, Serialize};
 use std::cmp;
 use std::fmt;
 use std::ops;
 
 /// A vector of bytes.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
```
udoprog, line 15: This should use a custom `Serialize` impl, since it can make use of [`Serialize::serialize_bytes`](https://docs.rs/serde/1.0.130/serde/trait.Serializer.html#tymethod.serialize_bytes) which for many formats can be dealt with more efficiently than sequences.

udoprog:
Marked as bug since this fixes a serialization issue with `bincode` since this is what's being used in rune-cli as well (!)

udoprog:
To mention briefly the distinction between `Value`, `ConstValue` and `IrValue`.
* `Value` is the runtime representation of any value in Rune. It is what's being stored on the stack.
* `ConstValue` is the subset of `Value` which 1) has no runtime dependencies and 2) is thread safe. It doesn't require access to a `Unit` nor access to a `Context`. Since it doesn't interact with the stack the values do not have to be cloned and are largely used when constants are needed. It is also used in `SyncFunction` to convert any values which are potentially captured by closures since `ConstValue` in contrast to `Value` is `Send` and `Sync`.
* `IrValue` can be ignored. It's only used as an intermediate value during constant evaluation with `IrInterpreter`. It's almost a perfect subset of `Value`, except it has a lossless integer type since there's no worries about performance here. It should hopefully be removed in favor of using `Value` directly in the next compiler version with a better intermediate interpretation!

crates/rune/src/runtime/bytes.rs
tgolsson: @udoprog Fixed!

crates/rune/src/runtime/bytes.rs
```diff
@@ -6,13 +6,15 @@ use crate::compile::{InstallWith, Named};
...
     FromValue, Mut, RawMut, RawRef, RawStr, Ref, UnsafeFromValue, Value, VmError,
 };
+use serde::{Deserialize, Serialize};
 use std::cmp;
 use std::fmt;
 use std::ops;
 
 /// A vector of bytes.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize)]
```
udoprog, line 15: Since `Bytes` is purely a container of bytes it's more appropriate that it gets serialized transparently as an array of bytes, rather than deriving `Serialize` on the struct which has a field `inner` that is serialized as bytes.

A way to achieve this would be to implement `Serialize` / `Deserialize` directly and call the appropriate serialization / deserialization functions. Alternatively I think you can add `#[serde(transparent)]` to this struct.

PR #317, opened 2021-11-15
store and expose public constants in the unit (fixes #316)

### LINKED ISSUE #316: Can't lookup constant values defined on the Rune side in Rust
test.rn:

```rust
pub const LOAD_COUNT = 1000000;
```

If I try to build it, first I'm getting a warning:

```
  ┌─ workloads/test.rn:1:1
  │
1 │ pub const LOAD_COUNT = 1000000;
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not used

```

And then when I try to lookup the value of the constant from the Rust program I'm getting a `None`:

```rust
let const_value = unit.constant(rune::Hash::type_hash(&["LOAD_COUNT"]));       // evals to None
```

Even making a faux reference to `LOAD_COUNT` in Rune sources fixes only the warning, but doesn't fix the lookup.

Rune version: 0.10.0  hash 331a4648

tgolsson:
@pkolaczk Please give this branch a try! There's a new `Hash::constant` that can be used to make it clearer, as well.

### DIFF
--- a/crates/rune-cli/src/main.rs
+++ b/crates/rune-cli/src/main.rs
@@ -628,7 +628,7 @@ async fn run_path(args: &Args, options: &Options, path: &Path) -> Result<ExitCod
                     }
                 }
 
-                if constants.peek().is_some() {
+                if runargs.dump_constants && constants.peek().is_some() {
                     writeln!(out, "# constants")?;
 
                     for constant in constants {
--- a/crates/rune-cli/src/main.rs
+++ b/crates/rune-cli/src/main.rs
@@ -95,6 +95,7 @@ impl Command {
             }
             Command::Run(args) => {
                 if args.dump {
+                    args.dump_constants = true;
                     args.dump_unit = true;
                     args.dump_stack = true;
                     args.dump_functions = true;
@@ -112,6 +113,7 @@ impl Command {
                     || args.dump_native_functions
                     || args.dump_stack
                     || args.dump_types
+                    || args.dump_constants
                     || args.emit_instructions
                 {
                     args.dump_unit = true;
@@ -227,6 +229,9 @@ struct RunFlags {
     /// Dump default information about unit.
     #[structopt(long)]
     dump_unit: bool,
+    /// Dump constants from the unit.
+    #[structopt(long)]
+    dump_constants: bool,
     /// Dump unit instructions.
     #[structopt(long)]
     emit_instructions: bool,
@@ -599,6 +604,7 @@ async fn run_path(args: &Args, options: &Options, path: &Path) -> Result<ExitCod
                 let mut functions = unit.iter_functions().peekable();
                 let mut strings = unit.iter_static_strings().peekable();
                 let mut keys = unit.iter_static_object_keys().peekable();
+                let mut constants = unit.iter_constants().peekable();
 
                 if runargs.dump_functions && functions.peek().is_some() {
                     writeln!(out, "# dynamic functions")?;
@@ -619,6 +625,14 @@ async fn run_path(args: &Args, options: &Options, path: &Path) -> Result<ExitCod
 
                     for string in strings {
                         writeln!(out, "{} = {:?}", string.hash(), string)?;
+                    }
+                }
+
+                if constants.peek().is_some() {
+                    writeln!(out, "# constants")?;
+
+                    for constant in constants {
+                        writeln!(out, "{} = {:?}", constant.0, constant.1)?;
                     }
                 }
 
--- a/crates/rune/src/compile/unit_builder.rs
+++ b/crates/rune/src/compile/unit_builder.rs
@@ -142,7 +142,6 @@ impl UnitBuilder {
 
             match self.constants.get(&to) {
                 Some(value) => {
-                    eprintln!("xxx");
                     let const_value = value.clone();
                     if self.constants.insert(from, const_value).is_some() {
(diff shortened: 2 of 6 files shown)


tests/tests/unit_constants.rs
```diff
@@ -0,0 +1,25 @@
+use rune::Hash;
+
+#[test]
+fn test_get_const() -> rune::Result<()> {
+    let context = rune_modules::default_context()?;
+
+    let mut sources = rune::sources! {
+        entry => {
+        pub const LOAD_COUNT = 1337;
```
udoprog, line 9: Also want to test for these (since I'm not sure they work right now!):
* A constant exported in a nested public module.
* A constant re-exported (through `pub use ...` from a nested private module.

crates/rune-cli/src/main.rs
```diff
@@ -622,6 +628,14 @@ async fn run_path(args: &Args, options: &Options, path: &Path) -> Result<ExitCod
                     }
                 }
 
+                if constants.peek().is_some() {
```
udoprog, line 631: Should check `dump_constants`?

```suggestion
                if runargs.dump_constants && constants.peek().is_some() {
```

crates/rune/src/compile/unit_builder.rs
```diff
@@ -495,10 +495,14 @@ impl UnitBuilder {
...
                 );
             }
+
             MetaKind::Function { .. } => (),
             MetaKind::Closure { .. } => (),
             MetaKind::AsyncBlock { .. } => (),
-            MetaKind::Const { .. } => (),
+            MetaKind::Const { const_value } => {
+                self.constants
+                    .insert(Hash::constant(&meta.item.item.to_string()), const_value.clone());
```
udoprog, line 504: Hm. I think this needs to use a type path so that nested constants are supported?

tests/tests/unit_constants.rs
tgolsson: Re-exports only work for functions but I'll fix that :D

crates/rune-cli/src/main.rs
tgolsson: Good point

crates/rune/src/compile/unit_builder.rs
tgolsson: Yepp! It's a bit weird to use `type_hash` here which feels weird, but yes. Maybe it should be `path`?

udoprog: Since we are renaming things - go for it! It might even make sense to rename the `Item` type into `Path` even if it might be minorly misconstrued as a filesystem path.

tgolsson: QualifiedName? :-)

tgolsson: (but renaming anything goes into another PR -- I'll keep type_hash for now and remove Hash::constant.)

crates/rune/src/compile/unit_builder.rs
```diff
@@ -126,22 +126,41 @@ impl UnitBuilder {
...
 
-            if self.functions.insert(from, info).is_some() {
-                return Err(CompileError::new(
-                    span,
-                    CompileErrorKind::FunctionConflictHash { hash: from },
-                ));
-            }
+            match self.constants.get(&to) {
+                Some(value) => {
+                    eprintln!("xxx");
```
udoprog, line 145: rogue statement
```suggestion
```

crates/rune/src/compile/unit_builder.rs
```diff
@@ -126,22 +126,41 @@ impl UnitBuilder {
...
+                            span,
+                            CompileErrorKind::ConstantConflict {
+                                item: Item::with_item(&["unknown"]),
+                                hash: from,
+                            },
+                        ));
+                    }
+                    continue;
+                }
+                None => {}
```
udoprog, line 158: Nit, but since there's only one branch in use this could do:
```rust
if let Some() = self.constants.get(&to) {
   /* .. */
}
```

crates/rune/src/compile/unit_builder.rs
```diff
@@ -126,22 +126,41 @@ impl UnitBuilder {
...
+                    let info = *info;
+                    if self.functions.insert(from, info).is_some() {
+                        return Err(CompileError::new(
+                            span,
+                            CompileErrorKind::FunctionConflictHash { hash: from },
+                        ));
+                    }
+                    continue;
                 }
+                None => {}
```
udoprog, line 140: Same here (see below)

tests/tests/unit_constants.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+#[test]
+fn test_get_const_re_export() -> rune::Result<()> {
+    let context = rune_modules::default_context()?;
+
+    let mut sources = rune::sources! {
+        entry => {
+            mod inner {
+		pub const LEET = 1337;
```
udoprog, line 34: Accidental tab (and a few more places)?

udoprog (approved):

pkolaczk:
You've just closed it when I was testing! No worries, works like a charm! 
Thank you! :tada:

udoprog:
Thanks for fixing this @tgolsson

@pkolaczk thanks for taking it for a spin!

PR #327, opened 2021-11-21
Add template literals in an empty group (fixes #326)

### LINKED ISSUE #326: Instance Functions on Template Literals Fail
Environment
* Arch Linux - Kernel 5.15.2
* Rust 1.56.1 - 2021 Edition
* Rune 0.10.0 (main branch) - commit fcf3fddccb8e87b9ebe24f0114625d8471e52e4d (latest commit as of a few days ago)

I've been evaluating Rune for a project I'm currently in the exploratory phase for and I came across an interesting issue. Given the following function being added to the runtime as an instance function:
```rust
fn trim_indent(string: String) -> String {
    string
        .lines()
        .map(|s| s.trim_start())
        .collect::<Vec<&str>>()
        .join("\n")
}
```
when trying to invoke it directly on a template literal it fails at runtime:
```rust
let template_runtime_failure = 
    `Hello, I am ${age} years old.
     How old are you?`.trim_indent();
```
However, after calling the function on it after initial assignment it behaves as expected:
```rust
let template_runtime_sucess = 
    `Hello, I am ${age} years old.
     How old are you?`;

let template_runtime_success = template_runtime_success.trim_indent();
```
I would  assume that any custom instance functions that would work on String would also work on template literals as  they do in Javascript. I'm guessing the runtime failure is due to the method being called on the template literal before it is evaluated.

Here is a full example for reproduction purposes:
```rust
use rune::{ContextError, Diagnostics, Module, Source, Sources, Vm};
use std::sync::Arc;

fn main() -> rune::Result<()> {
    let mut context = rune_modules::default_context()?;
    context.install(&trim_module()?)?;

    let runtime = Arc::new(context.runtime());

    let mut sources = Sources::new();
    sources.insert(Source::new(
        "script",
        r#"
        pub fn test_multiline_template() {
            let age = 35;
            
            let template_runtime_failure = 
                `Hello, I am ${age} years old.
                  How old are you?`.trim_indent();

            println(template_runtime_failure);
        }
        "#,
    ));

    let mut diagnostics = Diagnostics::new();
    
    let result = rune::prepare(&mut sources)
        .with_context(&context)
        .with_diagnostics(&mut diagnostics)
        .build();

    let unit = result?;
    let mut vm = Vm::new(runtime, Arc::new(unit));

    vm.call(&["test_multiline_template"], ())?;

    Ok(())
}

fn trim_module() -> Result<Module, ContextError> {
    let mut m = Module::with_item(&["mymodule"]);
    m.inst_fn("trim_indent", trim_indent)?;
    Ok(m)
}

fn trim_indent(string: String) -> String {
    string
        .lines()
        .map(|s| s.trim_start())
        .collect::<Vec<&str>>()
        .join("\n")
}
```

udoprog:
This causes template literal desugaring (which happens during lexing) to be desugared into an empty group so that attributes are associated with the internal `template!` macro call rather than the chained expression.

Thanks @Steven0351 for finding and reporting the issue!

### DIFF
--- a/crates/rune/src/ast/expr.rs
+++ b/crates/rune/src/ast/expr.rs
@@ -248,6 +248,14 @@ impl Expr {
         false
     }
 
+    /// Try to coerce into item if applicable.
+    pub(crate) fn into_item(self) -> Result<ast::Item, Self> {
+        match self {
+            Self::MacroCall(e) => Ok(ast::Item::MacroCall(e)),
+            e => Err(e),
+        }
+    }
+
     /// Parse an expression without an eager brace.
     ///
     /// This is used to solve a syntax ambiguity when parsing expressions that
@@ -357,15 +365,9 @@ impl Expr {
     pub(crate) fn parse_with_meta(
         p: &mut Parser<'_>,
         attributes: &mut Vec<ast::Attribute>,
-        path: Option<ast::Path>,
         callable: Callable,
     ) -> Result<Self, ParseError> {
-        let lhs = if let Some(path) = path {
-            Self::parse_with_meta_path(p, attributes, path, EagerBrace(true))?
-        } else {
-            Self::parse_base(p, attributes, EagerBrace(true))?
-        };
-
+        let lhs = Self::parse_base(p, attributes, EagerBrace(true))?;
         let lhs = Self::parse_chain(p, lhs, callable)?;
         Self::parse_binary(p, lhs, 0, EagerBrace(true))
     }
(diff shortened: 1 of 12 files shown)


udoprog:
A similar but different issue (since they're not desugared during lexing) *might* apply to other macros relying on `#[builtin]`. Like `format!(..)` and `line!()`.

udoprog:
Macro calls in items and statements *were* affected by a parsing bug. But a different one from this. It stems in how statements tries to "pre parse" a macro call which prevents it from being parsed as an expression chain.

The solution was to parse the statement as an expression an then try to lift it into an item using the `ast::Expr::into_item` function introduced here.

PR #345, opened 2021-11-29
Explicitly drop stack guards to avoid use-after-free issue

### LINKED ISSUE #344: Passing an interpolated string from Rune script to a Rust function passes garbage
I've got a Rune script:

```rust
const KEYSPACE = "latte";
const TABLE = "basic";

pub async fn schema(db) {
    db.execute(`CREATE KEYSPACE IF NOT EXISTS ${KEYSPACE} \
                    WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }`).await?;
    db.execute(`CREATE TABLE IF NOT EXISTS ${KEYSPACE}.${TABLE}(id bigint PRIMARY KEY)`).await?;
}

pub async fn erase(db) {
    println(`TRUNCATE TABLE ${KEYSPACE}.${TABLE}`);
    db.execute(`TRUNCATE TABLE ${KEYSPACE}.${TABLE}`).await
}
```

The execute instance function looks like this:

```rust
impl Session {
pub async fn execute(&self, cql: &str) -> Result<(), CassError> {
        println!("Executing: {}", cql);
        let statement = Statement::new(cql, 0);
        let start_time = self.stats.borrow_mut().start_request();
        let rs = self.inner.execute(&statement).await;
        let duration = Instant::now() - start_time;
        self.stats.borrow_mut().complete_request(duration, &rs);
        rs?;
        Ok(())
    }
}
```

and is registered this way:

```rust
 let mut session_module = Module::default();
        session_module.ty::<Session>().unwrap();
        session_module
            .async_inst_fn("execute", Session::execute)
            .unwrap();
```

Now, when I run this on Rune until commit 5fcf6e19bed6d5670aa2f2bfebfdadab120bccef, everything is fine and I'm getting:

```
info: Creating schema...
Executing: CREATE KEYSPACE IF NOT EXISTS latte WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }
Executing: CREATE TABLE IF NOT EXISTS latte.basic(id bigint PRIMARY KEY)
info: Erasing data...
TRUNCATE TABLE latte.basic
Executing: TRUNCATE TABLE latte.basic
```

However since commit 0fca7d6db68ee64486247d5b4c8cc1050567e9a3 the string passed to execute becomes garbage:

```
info: Creating schema...
Executing: CREATE KEYSPACE IF NOT EXISTS latte WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }
Executing: CREATE TABLE IF NOT EXISTS latte.basic(id bigint PRIMARY KEY)
info: Erasing data...
TRUNCATE TABLE latte.basic
Executing: ��
```

And this happens always in the call to `erase`, not call to `schema`. 
The garbage passed to execute is different each time. 
Reproduces always, both in debug and release mode.

If I replace string interpolation with a non-interpolated string (non template) all works fine (except it doesn't do what I want). 

On the Rust side, `Vm` is created fresh for each call, however it shares the `Context` and the `Unit` (only Arc ref is cloned).

udoprog:
CC: @pkolaczk to give this a try.

Hopefully fixes #344

I'm trying to come up with a good test for this before merging.

### DIFF
--- a/crates/rune/src/compile/module.rs
+++ b/crates/rune/src/compile/module.rs
@@ -829,8 +829,9 @@ macro_rules! impl_register {
                 #[allow(unused)]
                 let ret = unsafe {
                     impl_register!{@unsafe-vars $count, $($ty, $var, $num,)*}
-
-                    self($(<$ty>::unsafe_coerce($var.0),)*)
+                    let ret = self($(<$ty>::unsafe_coerce($var.0),)*);
+                    impl_register!{@drop-stack-guards $($var),*}
+                    ret
                 };
 
                 impl_register!{@return stack, ret, Return}
@@ -871,6 +872,7 @@ macro_rules! impl_register {
 
                     Future::new(async move {
                         let output = fut.await;
+                        impl_register!{@drop-stack-guards $($var),*}
                         let value = output.to_value()?;
                         Ok(value)
                     })
@@ -920,7 +922,9 @@ macro_rules! impl_register {
                 #[allow(unused)]
                 let ret = unsafe {
                     impl_register!{@unsafe-inst-vars inst, $count, $($ty, $var, $num,)*}
-                    self(Instance::unsafe_coerce(inst.0), $(<$ty>::unsafe_coerce($var.0),)*)
+                    let ret = self(Instance::unsafe_coerce(inst.0), $(<$ty>::unsafe_coerce($var.0),)*);
+                    impl_register!{@drop-stack-guards inst, $($var),*}
+                    ret
                 };
 
                 impl_register!{@return stack, ret, Return}
@@ -972,6 +976,7 @@ macro_rules! impl_register {
 
                     Future::new(async move {
                         let output = fut.await;
+                        impl_register!{@drop-stack-guards inst, $($var),*}
                         let value = output.to_value()?;
                         Ok(value)
                     })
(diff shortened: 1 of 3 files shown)


tgolsson (approved):
LGTM, but not sure about testing. Maybe running mir somewhere? If it works on Rune... Seems hard to catch a bug that is UB...

udoprog:
6dd7f6823466712cef47ea546e0d0c932a59e3bf introduces a test case. I rebased it to before the patch to ensure that it was indeed broken, and got this:

```
> git checkout 6dd7f68
> cargo test -- bug_344
running 4 tests
test bug_344::bug_344_function ... ok
test bug_344::bug_344_inst_fn ... ok
test bug_344::bug_344_async_function ... FAILED
test bug_344::bug_344_async_inst_fn ... FAILED

failures:

---- bug_344::bug_344_async_function stdout ----
thread 'bug_344::bug_344_async_function' panicked at 'value has was previously dropped: async argument', \\?\D:\Repo\rune\tests\tests\bug_344.rs:156:9

---- bug_344::bug_344_async_inst_fn stdout ----
thread 'bug_344::bug_344_async_inst_fn' panicked at 'value has was previously dropped: async self argument', \\?\D:\Repo\rune\tests\tests\bug_344.rs:156:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    bug_344::bug_344_async_function
    bug_344::bug_344_async_inst_fn

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 220 filtered out; finished in 0.00s 
```

This test then passed by df4a5ca:

```
> git checkout use-after-free
> cargo test -- bug_344
running 4 tests
test bug_344::bug_344_inst_fn ... ok
test bug_344::bug_344_function ... ok
test bug_344::bug_344_async_function ... ok
test bug_344::bug_344_async_inst_fn ... ok
```

PR #359, opened 2021-12-03
Start working on manifest loading

udoprog:
This adds very preliminary support for loading manifests from a `Rune.toml` file in a directory.

This should (eventually) allow for complex project structures such as those permitted through a `Cargo.toml` file in rust.

Right now, what is supported are the following:

```toml
[package]
name = "hello"
version = "0.0.0"

[workspace]
members = [
    "crates/a",
    "crates/b",
]
```

And what rune-cli does is to use this directory structure to autodiscover the following directories:
* `bin` for binary files that can be run with `rune run --bin <name>`.
* `tests` for files expected to contain test cases. If the `--test <name>` argument is omitted it runs all benchmarks by default.
* `examples` for examples that can be run with `rune run --example <name>`.
* `benches` for benchmarks that can be rune with `rune bench --bench <name>`. If the `--bench <name>` argument is omitted it runs all benchmarks by default.

# To do before merging

* [ ] The workspace is currently not validated. So for example multiple packages can have the same name. This is eventually important for when `rune build` is added, which is intended to produce a compacted bytecode package of all rune code with dependencies.
* [x] Wildcards in `members` are not supported.

### DIFF
--- a/Rune.toml
+++ b/Rune.toml
@@ -1,0 +1,5 @@
+[workspace]
+members = [
+    "benches",
+    "examples",
+]
--- a/benches/Rune.toml
+++ b/benches/Rune.toml
@@ -1,0 +1,3 @@
+[package]
+name = "rune-benches"
+version = "0.0.0"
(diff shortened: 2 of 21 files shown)


udoprog:
Merging this as the very first step of manifest loading. Even though it's fairly incomplete. Since it also touches a few (non-intrusive) improvements in rune-cli I don't want it to block anyone.

PR #369, opened 2021-12-09
add capture-io feature for rune-modules (fixes #368)

### LINKED ISSUE #368: Rune-cli fails to install from crates.io
Haven't dug too much yet. Everthing on latest stable.

Running `cargo install rune-cli` fails with the below error:

image

tgolsson:


udoprog (approved):
It's very unfortunate that the local build works by default *because* one of the other modules (`rune-wasm`) add a dependency for `capture-io`. I'd really want to get some build test for this. For now it might be sufficient to add a separate build step which builds `rune-cli`.

Thanks for the fix!

PR #371, opened 2021-12-09
Add Clone to Unit and Context

pkolaczk:
On multiprocessor computers, sharing the
same Unit (or Context) behind a single Arc can
be very costly, because the Arc(s) used inside
those structures end up in the
caches of all cores, and each update on
one core invalidates caches of the other cores.

By allowing Unit and Context to be Clone,
the users are able to do deep per-thread copies,
effectively unsharing all the reference counters
and letting each core progress at full
speed independently, at the expense of higher memory use.

### DIFF
--- a/crates/rune/src/runtime/debug.rs
+++ b/crates/rune/src/runtime/debug.rs
@@ -9,7 +9,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 
 /// Debug information about a unit.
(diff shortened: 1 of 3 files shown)


udoprog:
Does `Context` need to be `Clone` as well? It's a fairly complex data structure intended only for compilation, so cloning should be discouraged.

pkolaczk:
Removed `Clone` from `Context`.

udoprog:
Much obliged!

Issue #390, opened 2022-01-20
The INDEX_SET protocol fn now expects a unit return


jrobsonchase:
I updated to `main` to take advantage of the new enum goodies, and this broke unexpectedly.

Using a previously-accepted function that returns an `Option` as the `INDEX_SET` function now causes a vm error: 
```
Error: expected `unit`, but found `Option`
```

This is currently broken in the implementation of `HashMap` in `crates/rune/src/modules/collections.rs`. Also opening a PR adding a test case illustrating the failure.

udoprog:
I'll triage this properly when I get back home. But for now, INDEX_SET should probably require that a unit is returned unless there's a good reason not to. But that would mean a major version bump since the protocol is more restrictive.

PR #391, opened 2022-01-20
Fix HashMap INDEX_{SET,GET}

jrobsonchase:
Failing test case illustrating the breakage from #390.

If cherry-picked onto the 0.10.3 tag, it passes without issue.


udoprog:
Hey, Sorry for the late reply. I'm currently unavailable until the beginning of February. But I approved the run so we can see the results. I'll review when I get back home!

udoprog:
All right, so fixing the issue now. Cheers!

Issue #377, opened 2021-12-24
Can't parse float number in JSON


twigly:
Hi Rune Team,

The `Value` enum has a `Float` variant [here](https://github.com/rune-rs/rune/blob/main/crates/rune/src/runtime/value.rs#L235).
But there is no `visit_f32` and `visit_f64` methods in the `Visitor` trait implementation ([here](https://github.com/rune-rs/rune/blob/main/crates/rune/src/runtime/value.rs#L1361)).

As a result, it's not possible to parse JSON containing float numbers. The program panics on the the following error:
`"invalid type: floating point '2.1', expected any valid value"`

I'm happy to fix the issue if you like.

udoprog:
Nice find! This should be fixed.

PR #393, opened 2022-02-03
Add f32/f64 to VmVisitor

dranikpg:
I've just stumbled over [this issue](https://github.com/rune-rs/rune/issues/377) myself and took the time to fix.

The VmVisitor must have been copypasted from the KeyVisitor and the floats were left forgotten. 
A macro would actually seem to fit here to avoid all those duplicates.


udoprog:
Very nice. Thank you!

udoprog:
CI had an unrelated clippy failure. Cheers!

PR #401, opened 2022-06-09
add char::to_int

jrobsonchase:
There's a `char::from_int`, but I couldn't find a way to go the opposite direction.

Also adds some small tests for it.

### DIFF
--- a/crates/rune/src/modules/char.rs
+++ b/crates/rune/src/modules/char.rs
@@ -10,6 +10,7 @@ pub fn module() -> Result<Module, ContextError> {
     module.ty::<ParseCharError>()?;
 
     module.function(&["from_int"], char_from_int_impl)?;
+    module.function(&["to_int"], char_to_int_impl)?;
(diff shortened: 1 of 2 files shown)


crates/rune/src/modules/char.rs
```diff
@@ -23,6 +24,10 @@ pub fn module() -> Result<Module, ContextError> {
     Ok(module)
 }
 
+fn char_to_int_impl(value: char) -> Result<Value, VmError> {
+    Ok((value as i64).into())
```
udoprog, line 28: Not sure this will behave as intended on high glyphs. Or maybe wrapping when the unsigned value should be intended (just to follow the bit pattern). I'll accept this for now, but we might want to consider something different in the future for integer conversion.

udoprog:
Sorry this slipped through the cracks!

Sans some formatting and things we need to think about in the future (see comment) this looks good to me. Merging as-is!

PR #408, opened 2022-07-12
Emit doc attributes from doc comments

Xemiru:
When encountering comments, the lexer will now attempt to transform them into tokens to later be interpreted as doc attributes should they match the format of Rust's own style of doc comments, as described by [syn::Attribute's doc comments section](https://docs.rs/syn/1.0.98/syn/struct.Attribute.html#doc-comments).

### DIFF
--- a/crates/rune/src/parse/lexer.rs
+++ b/crates/rune/src/parse/lexer.rs
@@ -37,6 +37,46 @@ impl<'a> Lexer<'a> {
         self.iter.span_to_len(0)
     }
 
+    fn emit_doc_attribute(&mut self, inner: bool, span: Span, docstring_span: Span) {
+        // outer: #[doc = ...]
+        // inner: #![doc = ...]
+
+        self.buffer.push_back(ast::Token { kind: K![#], span });
+
+        if inner {
+            self.buffer.push_back(ast::Token { kind: K![!], span });
+        }
+
+        self.buffer.push_back(ast::Token {
+            kind: K!['['],
+            span,
+        });
+
+        self.buffer.push_back(ast::Token {
+            kind: ast::Kind::Ident(ast::LitSource::BuiltIn(ast::BuiltIn::Doc)),
+            span,
+        });
+
+        self.buffer.push_back(ast::Token {
+            kind: K![=],
+            span,
+        });
+
+        self.buffer.push_back(ast::Token {
+            kind: ast::Kind::Str(ast::StrSource::Text(ast::StrText {
+                source_id: self.source_id,
+                escaped: false,
+                wrapped: false,
+            })),
+            span: docstring_span,
+        });
+
+        self.buffer.push_back(ast::Token {
+            kind: K![']'],
+            span,
+        });
+    }
+
     fn emit_builtin_attribute(&mut self, span: Span) {
         self.buffer.push_back(ast::Token { kind: K![#], span });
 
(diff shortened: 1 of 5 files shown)


udoprog:
Looks great so far!

crates/rune/src/parse/lexer.rs
```diff
@@ -559,12 +599,72 @@ impl<'a> Lexer<'a> {
...
+                                    (false, false)
+                                }
+                            };
+
                             let term = self.consume_multiline_comment();
-                            break ast::Kind::MultilineComment(term);
+                            if !term {
+                                return Err(ParseError::new(
+                                    self.iter.span_to_pos(start),
+                                    ParseErrorKind::ExpectedMultilineCommentTerm,
```
udoprog, line 655: Long term I'd probably prefer if this error was emitted at a later stage, like during parsing. Since then the lexer is more resilient and can be re-used for parsers which are more suitable for live editing where comments are incomplete (future plan).

All though at the moment it makes little difference.

crates/rune/src/parse/lexer.rs
```diff
@@ -559,12 +599,72 @@ impl<'a> Lexer<'a> {
...
+                                let span = self.iter.span_to_pos(start);
+                                self.emit_doc_attribute(inner, span, span.clone().trim_start(3));
+                                continue 'outer;
+                            } else {
+                                break ast::Kind::Comment;
+                            }
                         }
                         ('/', '*') => {
+                            self.iter.next();
+                            let (doc, inner) = match self.iter.peek() {
```
udoprog, line 633: Consider breaking into fn.

crates/rune/src/parse/lexer.rs
```diff
@@ -559,12 +599,72 @@ impl<'a> Lexer<'a> {
                             break ast::Kind::PipeEq;
                         }
                         ('/', '/') => {
+                            self.iter.next();
+                            let (doc, inner) = match self.iter.peek() {
```
udoprog, line 603: Consider breaking into fn.

crates/rune/src/compile/meta.rs
```diff
@@ -363,6 +387,8 @@ pub(crate) struct ItemMeta {
     pub(crate) visibility: Visibility,
     /// The module associated with the item.
     pub(crate) module: Arc<ModMeta>,
+    /// Doc comment attributes, if any.
+    pub(crate) docs: Arc<Vec<Doc>>,
```
Xemiru, line 394: `const` items proved to be difficult when doc content was placed in `PrivMeta`. `ItemMeta` let them hold docs without too much friction, and seems to be a much better place for them to live, anyway.

tests/tests/compiler_docs.rs
```diff
@@ -0,0 +1,169 @@
...
+        "Module::Enum" => { " Module enum.\n" }
+        "Module::Enum::A" => { " Enum variant A.\n" }
+        "Module::Enum::B" => { " Enum variant B.\n" }
+    };
+
+    let mut sources = sources(r#"
+        //! Mod/crate doc.
+        /*! Multiline mod/crate doc.
+         *
+         *  Both of these comments disappear; we can't hold onto file-level attributes at the moment..
```
Xemiru, line 113: [New issue?](https://github.com/rune-rs/rune/blob/0ed2e7db04b20206c3cb18f0ccac2a63da16c287/crates/rune/src/indexing/index.rs#L585)

crates/rune/src/compile/compile_visitor.rs
```diff
@@ -15,6 +15,13 @@ pub trait CompileVisitor {
 
     /// Visit something that is a module.
     fn visit_mod(&mut self, _source_id: SourceId, _span: Span) {}
+
+    /// Visit an item's doc comment attribute.
+    ///
+    /// This may be called several times for a single item. Each attribute should eventually be
+    /// combined for the full doc string.
+    fn visit_doc_comment(&mut self, _source_id: SourceId, _meta: MetaRef<'_>, _span: Span, _docstr: &str) {}
```
Xemiru: Might be worth rewriting this in such a way where its called once with a collection of all doc comment attributes instead of being called for every attribute (span points to the whole set of comments and `docstr` becomes a `&Vec[&str]`).

Added benefit of immediately knowing how many comments there are for an item; this currently would have you only know at the end of compilation where you know that the function can't be called anymore.

tests/tests/compiler_docs.rs
```diff
@@ -0,0 +1,169 @@
...
+        "Struct" => {
+            " Top-level struct.\n"
+            " Second line!\n"
+        }
+        "Enum" => { "\n         * Top-level enum.\n         " }
+        "Enum::A" => { " Enum variant A.\n" }
+        "Enum::B" => { " Enum variant B.\n" }
+        "constant" => { " Top-level constant.\n" }
+
+        // Curiously, the module item itself never get registered. This always fails.
```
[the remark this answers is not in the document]
udoprog: See comment about files. Modules suffer from two issues that need to be addressed:

* We currently don't treat these as meta items (e.g. module information is not required for compilation) so we don't call the associated visitors.
* Interior doc comments require that we traverse the inside of the module and collect those comments somehow.

Both of these require minor changes. The easiest might be to simply make sure modules register `PrivMeta` items. This would even to provide [better diagnostics](https://rune-rs.github.io/play/?c=bW9kIHRlc3Qgewp9CgpwdWIgZm4gbWFpbigpIHsKICAgIHRlc3QoKQp9Cg%3D%3D) during compilation.

So instead of:
```
error: compile error
  ┌─ entry:5:5
  │
5 │     test()
  │     ^^^^ missing item `test`
```
We could have something like:
```
error: compile error
  ┌─ entry:5:5
  │
5 │     test()
  │     ^^^^ `test` is a module which cannot be called
```

tests/tests/compiler_docs.rs
udoprog: Yeah, this PR is in a decent state so lets break it into a separate one.

Capturing file and module level documentation requires adding calls to the associated visitor.

crates/rune/src/compile/meta.rs
udoprog: Might revisit later exactly how it's done (e.g. maybe we should prefer storing docs indexed separately in `Query`) but it's good enough for now.

crates/rune/src/compile/compile_visitor.rs
udoprog: Not the worst thing in the world, but if you discover a better approach go for it!

Issue #409, opened 2022-07-14
Collect docstrings for modules


udoprog:
Following #408, the way that docstring collection is currently hooked up does not emit docstrings for modules since it works at the layer where items emit `PrivMeta` and modules does not emit any.

This issue is opened to track docstring emission for modules (including the root module in files) since it requires a bit more work.

udoprog:
Closed in #410

PR #410, opened 2022-07-14
File and module docs

Xemiru:
A new meta type is introduced and registered by modules, allowing them to be picked up by the `CompileVisitor::register_meta` function just like any other item.

This also allows doc comments from modules to make their way to `CompileVisitor::visit_doc_comment` (#409). This does **not** include file-level root modules.

File-level doc comments are instead routed to a new function, `CompileVisitor::visit_file_doc_comment`.

### DIFF
--- a/crates/rune/src/query/mod.rs
+++ b/crates/rune/src/query/mod.rs
@@ -259,7 +259,7 @@ impl<'a> Query<'a> {
         span: Span,
         parent: &Arc<ModMeta>,
         visibility: Visibility,
-        docs: Arc<Vec<Doc>>,
+        docs: Arc<[Doc]>,
     ) -> Result<Arc<ModMeta>, QueryError> {
         let item = self.insert_new_item(items, source_id, span, parent, visibility, docs)?;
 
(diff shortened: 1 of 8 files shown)


crates/rune/src/query/mod.rs
```diff
@@ -273,6 +273,17 @@ impl<'a> Query<'a> {
             .modules
             .insert(item.item.clone(), query_mod.clone());
         self.insert_name(&item.item);
+        self.insert_meta(
```
Xemiru, line 276: There's a method, `CompileVisitor::visit_mod`, that seems to intend to be called when modules are encountered. However, the example in `compiler_docs.rs` doesn't seem to trigger it despite having multiple modules.

I'm unsure what cases it gets called in, but I'm wondering if this change obsoletes the method given that modules will now find themselves caught by `register_meta`.

udoprog: So `visit_mod` is intended to emit spans which "looks like a module". I don't think there's anything more to it. That's why it's being emitted everywhere that module-like expressions are encountered and not just where the module is defined.

Issue #413, opened 2022-07-15
Collect and expose metadata for struct and variant fields for docstrings


udoprog:
Fields do not have any `PrivMeta`, so their documentation strings are not being collected following #408 and therefore cannot be walked in #412 (and eventually used in generated documentation).

@Xemiru this might be an interesting issue for you to work on.

PR #415, opened 2022-07-16
Support documentation for fields in structs / struct variants.

Xemiru:
~~Adds an extra parameter to `insert_new_item` to receive doc comments for child objects contained within the item (targets #413).~~

~~This change assumes that we don't want to treat fields themselves as items, similar to how [Rust itself doesn't.](https://doc.rust-lang.org/reference/items.html)~~

Adds a new callback to `CompileVisitor` specifically for fields.

### DIFF
--- a/crates/rune-cli/src/doc.rs
+++ b/crates/rune-cli/src/doc.rs
@@ -75,7 +75,17 @@ fn walk_items(io: &mut Io<'_>, doc: &DocFinder, queue: &mut VecDeque<ItemBuf>) -
         writeln!(io.stdout, "module: {}", &current)?;
 
         for c in doc.structs.get(&current).into_iter().flatten() {
-            writeln!(io.stdout, "struct {}", c)?;
+            let struct_key = current.join(&[c.as_component_ref()]);
+            if let Some(docs) = doc.field_docs.get(&struct_key) {
+                writeln!(io.stdout, "struct {} {{", c)?;
+                for field in docs.keys() {
+                    writeln!(io.stdout, "  {}", field)?;
+                }
+
+                writeln!(io.stdout, "}}")?;
+            } else {
+                writeln!(io.stdout, "struct {}", c)?;
+            }
         }
 
         for c in doc.enums.get(&current).into_iter().flatten() {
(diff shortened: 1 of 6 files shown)


udoprog:
Hm. It might be better to have a separate callback for fields rather than synthesizing an item just to be able to use the same visitor. That would allow for avoiding allocating a separate collection of `child_docs` and the item itself.

Something like this added to `CompileVisitor`:
```rust
fn visit_field_doc_comment(location: Location, item: &Item, field: &str, doc: &str);
```

udoprog:
Also consider adding field listing to `rune doc` while you're at it so you get a feel for what works (and what doesn't)! It's helped me to discover what is and what isn't needed.

crates/rune-cli/src/doc.rs
```diff
@@ -103,6 +123,7 @@ fn walk_items(io: &mut Io<'_>, doc: &DocFinder, queue: &mut VecDeque<ItemBuf>) -
 struct DocFinder {
     meta: BTreeMap<ItemBuf, MetaKind>,
     docs: HashMap<ItemBuf, Vec<String>>,
+    field_docs: HashMap<ItemBuf, HashMap<Box<str>, Vec<String>>>,
```
Xemiru, line 126: Potential conflict when a variant's path and a struct's path happen to collide.

udoprog: Yeah, this should be disallowed on a language level, but currently it uh, does something weird (spins forever) so I just found a bug :):

```rust
enum Foo {
    /// Hello
    Variant,
}

mod Foo {
    //// Hello
    struct Variant;
}
```

In Rust, this would give and we should replicate it:

```
1 | enum Foo {
  | -------- previous definition of the type `Foo` here
...
6 | mod Foo {
  | ^^^^^^^ `Foo` redefined here
  |
  = note: `Foo` must be defined only once in the type namespace of this module
```

Unless you can think of another way this can conflict?

Xemiru: This setup properly crashes with the below message when rolling back to bd33c6b796f676242007fdc1d6bf0fb4559fbc13. Trying to narrow it down further, at the moment.

```
error: query error
   ┌─ main:24:13
   │    
24 │ ╭ ╭             struct Foo {
25 │ │ │                 /// Struct field A.
26 │ │ │                 a,
27 │ │ │                 /// Struct field B.
28 │ │ │                 b,
29 │ │ │             }
   │ ╰─│─────────────^ `Enum::Foo` can refer to multiple things
   │   ╰─────────────' here as `Enum::Foo`
   · │ │
33 │                 Foo {
   │                 --- here as `Enum::Foo`
```

Xemiru: And yeah, that was the way I thought it'd conflict. I wasn't able to reach the above crash and deem the conflict impossible at this stage due to the spinning forever thing.

Xemiru: Spins after 003a32b0990ddd7bd25b5d69e02071674c25c8c9.

udoprog: I think this was due to a change a while back which allows for multiple items to be indexed (to improve diagnostics), but eventually only one should be allowed to be inserted unambiguously.

The issue arises when the meta cache is queried before exhausting indexed entries, causing the loop that you identified on Discord to never exit. This should be fixed by #417.

udoprog:
Cheers!

PR #416, opened 2022-07-16
Track fields in struct metas.

Xemiru:
Introduces more information on fields in `PrivStructMeta` (targets #413).

Field information is largely similar to regular items; this is thus performed by storing `ItemMeta` data for the fields. They are not treated as actual items (the metas don't get registered), [similar to how Rust treats them.](https://doc.rust-lang.org/reference/items.html)

(is this still necessary after #414 / #415?)

### DIFF
--- a/crates/rune/src/compile/context.rs
+++ b/crates/rune/src/compile/context.rs
@@ -9,7 +9,7 @@ use crate::compile::module::{
     TypeSpecification, UnitType, VariantKind,
 };
 use crate::compile::{
-    ComponentRef, IntoComponent, Item, ItemBuf, Meta, Names, PrivMeta, PrivMetaKind,
+    ComponentRef, IntoComponent, Item, ItemBuf, ItemMeta, Meta, Names, PrivMeta, PrivMetaKind,
     PrivStructMeta, PrivTupleMeta, PrivVariantMeta,
 };
 use crate::runtime::{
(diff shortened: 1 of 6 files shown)


crates/rune/src/compile/context.rs
```diff
@@ -423,7 +423,22 @@ impl Context {
...
+                            .clone()
+                            .into_iter()
+                            .map(|it| {
+                                let mut item = item.clone();
+                                item.push(&it);
+                                (
+                                    it,
+                                    Arc::new(ItemMeta {
+                                        item,
+                                        ..ItemMeta::default()
```
Xemiru, line 437: Does more information need to be here?

crates/rune/src/compile/context.rs
```diff
@@ -442,7 +457,22 @@ impl Context {
...
+                                        .clone()
+                                        .into_iter()
+                                        .map(|it| {
+                                            let mut item = item.clone();
+                                            item.push(&it);
+                                            (
+                                                it,
+                                                Arc::new(ItemMeta {
+                                                    item,
+                                                    ..ItemMeta::default()
```
Xemiru, line 471: ditto

crates/rune/src/query/mod.rs
```diff
@@ -1633,18 +1633,30 @@ fn tuple_body_meta(
...
+        let name = field.name.resolve(ctx)?;
+        buf.push(name);
+
+        fields.insert(
+            name.into(),
+            Arc::new(ItemMeta {
+                id: Id::default(),
+                location: Location::new(item.location.source_id, field.span()),
+                item: buf,
+                visibility: Visibility::from_ast(&field.visibility).unwrap(),
```
Xemiru, line 1656: Need a way to avoid this unwrap.

Xemiru: This was replaced by using `QueryError::msg`, though I wonder if it's possible to even encounter this error at this stage.

udoprog:
Yeah, this might no longer be necessary since I opted for calling the visitor directly in #414.

PR #423, opened 2022-08-10
Fix language server

SWW13:
Currently the language server iterates over a previously cleared vector of source files, this renders the language server useless.

Only tested with my WiP rework of the vscode extension, but this fix should work with the current extension as well.


udoprog:
Thanks for looking at the language server! I've mostly been performing refactoring as dependencies are bumped and this was clearly missed!

I expect #425 to be merged as well once I've taken it for a spin with this fix in place.

Issue #407, opened 2022-07-06
Does the VSCode extension work as intended?


scambier:
Hello, I've installed the extension, and configured it to set the full path to the rune-languageserver.exe. However, it looks like it's doing nothing, and I have literally 0 log from "Rune (Client)" in the Output window.

By "doing nothing" I mean that I have a .rn file open in vscode, with a few errors that go undetected. Syntax highlighting works, but that's it.

udoprog:
Hm. Haven't tried it in a while, so not sure. I'll take it for a spin when I have a minute.

The vscode extension itself could use a bump. I think it's coded against a fairly old version of vscode by now, which shouldn't cause issues in theory but who knows?

udoprog:
If someone is curious to pick up the vscode extension I'd be happy to provide mentoring instructions.

GlenDC:
In case the project is meant to survive for more years to come including support, I would love to help out. This could be my first contribution. How can I help out?

udoprog:
@GlenDC Cool, you should join the Discord! The first step should be to get a development server up and running inside of vscode.

The following is a small guide on how to get up and running with vscode development:

The language server should "just work", by building and running it like this:
```
cargo run --bin rune-languageserver
```
The vscode extension itself is found in `editors/code`, and might do with some modernization. Following the [Your First Extension guide](https://code.visualstudio.com/api/get-started/your-first-extension) should give a good idea of how it should work. I just tried to open the project in vscode and run the extension, but I'm not getting a development window at this point so this might be the first thing that needs to be fixed (!).

When you get a development window up and running you need to find the binary path of the built server and configure this as the (`binaryPath` setting). This path should be should be `{project}/target/debug/rune-languageserver` where `{project}` is the location of the project.

At this point you should be running a development language server. When you rebuild the language server (`cargo build --package rune-languageserver`) all you should need to do is reload the development window `CTRL+P - Developer: Reload Window`.

udoprog:
Also note that I did try to build a completely new project using yeoman (`yo code` as shown in the tutorial above), and I got that to run. So porting the old extension to a completely new project structure might be a good way to move forward.

GlenDC:
[...]

PR #425, opened 2022-08-10
Rework vscode extension

SWW13:
Following the suggestion in #407 this is a complete rework of the vscode extension based on the rust-analyzer extension code.

This should at least fix the usage over ssh connection and other non-local as the deprecated `globalStoragePath` is no longer used. It also looks like the current extension is not working at all, at least for me. I won't investigate why.

In case someone wants to test the new extension (and the fixed `rune-languageserver`), I pre-built the binary (linux only) and the extension:
- [rune-languageserver](https://pixeldrain.com/u/BDe3BxP6)
- ~~[rune-vscode-0.12.0.vsix](https://pixeldrain.com/u/V5aPj6KY)~~ [rune-vscode-0.12.0.vsix](https://pixeldrain.com/u/uac75TPK)


udoprog:
> I won't investigate why.

Goooood call!

Once I've given it a run over to make sure everything works as expected on my end as well I expect to merge this straight up. Thank you for doing this!

udoprog:
This runs for me.

The only thing I was lacking was a launch configuration, so I took a basic one from the `yo code` project I set up. I'm also not very familiar with yarn so there was some confusion on my part where I used it and npm interchangeably which vscode extension development didn't like.

I also relaxed the vscode requirement from 1.70.0 to 1.66 which is the same as rust-analyzer uses, since that is where most of the code comes from it should be compatible.

One nit that I'd like to fix before this plugin is uploaded is that file icons uses the extension icon, which aren't designed to be icons and look kinda ugly. But this is not a blocker for merging.

image

Thank you for the work!

SWW13:
If you copied over the vscode launch configuration from `yo` you should make sure it's for a type script project with yarn and webpack enabled, otherwise the debugging isn't working as expected.

Also the `Rune: Restart server` command throws some errors and I don't know how to fix them, but it does restart the server - so better than nothing.

SWW13:
The file icon(s) are set in https://github.com/rune-rs/rune/blob/75e44cbbbefe00eab9771386bda8e8af34c3f033/editors/code/package.json#L42-L45 and can be removed.

udoprog:
> Also the `Rune: Restart server` command throws some errors and I don't know how to fix them, but it does restart the server - so better than nothing.

What's the error? Might be that the languageserver itself doesn't understand how to perform an orderly shutdown yet.

SWW13:
> Despite the `"type": "npm"` in the `tasks.json` [...]

Same for me, should be fine I guess.

> What's the error? Might be that the languageserver itself doesn't understand how to perform an orderly shutdown yet.

```
[Error - 14:14:08] Stopping server failed
  Message: Pending response rejected since connection got disposed
  Code: -32097
```
Probably due to the heavy adoption from the rust-analyzer code our extension does to much things different and the dispose of object got out of the expected order.

Issue #263, opened 2021-05-17
Overloadable Index


lupan:
This is 2 different things which I have used many times in lua for more powerful dynamic object handling, simply the act of overloading what happens when you use either `object.whatever` and `object["whatever"]`, I have used this many times to write dynamic message systems via internal templates to create fully runtime handled objects, this is horrible to use in rust due to fact its a lot of function calls, but in scripting langs it is quite simple and beautiful, for example:

`struct template { //for generating the msg at runtime
    name: String,
    fields: Vec<(String, TypeEnum)> //name and default value
}
struct dynamic_msg {
    template: Arc<template>,
    fields: HashMap<String, TypeEnum> //name of field and inner value (this is cloned from template)
}
impl dynamic_msg {
  fn set_dynamic_field(&mut self, name: &str, new_value: TypeEnum) -> Result<()> {
      self.fields[name] = new_value; //missed out the checks for simplicity sake
      Ok(())
  }
fn get_dynamic_field(&self, name: &str) -> Result<&TypeEnum> {
      let val = self.fields.get(name); //missed out the checks for simplicity sake
      Ok(val)
  }
fn get_dynamic_field_mut(&mut self, name: &str) -> Result<&mut TypeEnum> {
      let val = self.fields.get_mut(name); //missed out the checks for simplicity sake
      Ok(val)
  }
}`

in lua I usually override the index function on the userdata and make it simply run the set and get dynamic field functions, allowing me in lua to write this:
`msg = create_msg("name")
msg.field = "potato"
etc etc
return msg`

I realize this is possible using Object but that comes with 2 issues for me, one it doesn't handle if I try to access or create a field which isn't part of the msg, and it has to be parsed from and into the Object type, as for the dynamic struct system which is great, it requires the scripter to write the structs themselves.

p.s. this example above is purposely simple and far from what the optimized and expanded version would look like.

LoopyAshy:
I actually agree with this; but only for the fact that `Objects` can already access their fields both via [name] and .name, overall this being a option on the `INDEX_GET`/`INDEX_SET` protocols would be wonderful. Lua indeed allows this feature so I am quite used to it and overall it makes the code much nicer to read and write along with maintaining safety, which is what a scripting language should be.

udoprog:
I'm not entirely following how this is different from the `INDEX_SET` and `SET` protocols (e.g. the former is `object[key]`, the latter `object.key`).

[...]

PR #440, opened 2022-10-19
Attempt FALLBACK_SET/GET on failure to find field via SET/GET

### LINKED ISSUE #381: Object Function Overloading
Maybe Add to Module to have Object call overrides than will run a set function for checks and returns similar to **module.field_fn** but as **object_fn**.

So maybe `module.object_fn(Type_Npc , "obj", Npc::handle_object_lookup)?;`

`pub fn handle_object_lookup(&mut self, key: String) -> Value {}`

```
struct Type_Npc {
  obj: rune::runtime::Object,
}

```
So when someone calls 

`obj.x`

then x will be sent as the key for lookup and return. this way we could lookup data and error it out if it doesn't Exist on the script side which would allow for slightly more dynamic structures. OR static Data returns that don't need to be pre-made or saved into the struct.

As request by "Granny Joe" on Discord.  For more information or if I'm a bit wrong on what they are asking Ill link them here so they can fix anything missed in comments.

LoopyAshy:
I made it possible to fallback to ~~`Protocol::INDEX_GET`~~`Protocol::FALLBACK_GET` with `StaticString`/`String` key on failure to find a field via `Protocol::GET`, the same applies to ~~`Protocol::INDEX_SET`~~`Protocol::FALLBACK_SET` and `Protocol::SET`.

## Reasoning
Sometimes it is desired for user readability to pretend a specific custom type has fields and instead accessing the data from elsewhere such as from a `HashMap` or another collection.
possible use cases is having types with dynamic layouts while having rules to their usage such as locking the layout once created.
 it also allows custom types to be dynamically extended by the scripter while keeping the benefits of being a static `Shared<AnyObj>` but not seeming like a collection which `[]` accesses give the impression of. 

## Benchmarks
**Note**
Due to the tiny footprints of the functions, the differences are very hard to see if any and can be improved with a faster hashmap regardless  but honestly values these small are easily within margin of error.

### Before
```
test fields::actual_field_get                    ... bench:         418 ns/iter (+/- 88)
test fields::actual_field_set                    ... bench:         434 ns/iter (+/- 161)

test fields::string_key_classic_index_get        ... bench:         523 ns/iter (+/- 52)
test fields::static_string_key_classic_index_get ... bench:         486 ns/iter (+/- 80)
```
### After
```
test fields::actual_field_get                     ... bench:         413 ns/iter (+/- 47)
test fields::actual_field_set                     ... bench:         432 ns/iter (+/- 25)

test fields::string_key_classic_index_get         ... bench:         500 ns/iter (+/- 14)
test fields::static_string_key_classic_index_get  ... bench:         469 ns/iter (+/- 32)

test fields::string_key_fallback_field_get        ... bench:         499 ns/iter (+/- 15)
test fields::static_string_key_fallback_field_get ... bench:         469 ns/iter (+/- 20)

test fields::string_fallback_field_set            ... bench:         596 ns/iter (+/- 24)
test fields::static_string_fallback_field_set     ... bench:         608 ns/iter (+/- 14)
```

## Closes
#381 #263 

## Notes
Regarding the changes done to this PR, I do wonder if FALLBACK_* is preferred choice compared to META_* or META_FIELD_* since in most script languages I have used, META usually implies it does it first/instead not after, which although being a option would be a fine choice.

I do like the advantage of the current design being that it prioritises actual fields first.

I do wonder if it was changed to use the 'meta' getter/setter first if it would be plausible to pass in the original field getter and setter functions so u could do this:
```rust
|this: &Custom, default_get: NativeFunction, name: String| if let Ok(v) = default_get(name) {
    Ok(v)
} else {
    //Do my field get
}
```
This approach however would have overhead and although could be faster for types which only want 'meta' fields I definitely feel it would cost for classic field get/sets which would be a waste and a shame and it isn't the most attractive solution regardless.

Another possible solution could be explicitly declaring a type as having 'meta' fields and which order to prioritise finding them such as `MetaFieldAccess::First` `MetaFieldAccess::Last`, `MetaFieldAccess::Always` and finally `MetaFieldAccess::Never`(default) which would be drastically cheaper and simpler than it being passed in as a argument to the meta function which in my opinion reduces the understandability for implementors.
The enum system could even be applied to the `Any` derive macro as an attribute to make it easy to implement and recognise by looking at types.

### DIFF
--- a/crates/rune/src/runtime/vm.rs
+++ b/crates/rune/src/runtime/vm.rs
@@ -533,28 +533,6 @@ impl Vm {
         Ok(CallResult::Unsupported(target))
     }
 
-    /// Helper to call a field function which has been prechecked to exist.
-    #[inline(always)]
-    fn call_field_fn_prechecked<A>(
-        &mut self,
-        target: Value,
-        handler: &Arc<dyn Fn(&mut Stack, usize) -> Result<(), VmError> + Send + Sync>,
-        args: A,
-    ) -> Result<CallResult<()>, VmError>
-    where
-        A: Args,
-    {
-        let count = args.count();
-        let full_count = count + 1;
-
-        self.stack.push(target);
-        args.into_stack(&mut self.stack)?;
-
-        handler(&mut self.stack, full_count)?;
-
-        Ok(CallResult::Ok(()))
-    }
-
     /// Helper to call an index function.
     #[inline(always)]
     fn call_index_fn<A>(
@@ -948,28 +926,16 @@ impl Vm {
                 }
             }
             target => {
-                let hash = index.hash();
-                let type_hash =
-                    Hash::field_fn(Protocol::GET, target.type_hash()?, hash.into_type_hash());
-                return Ok(if let Some(handler) = self.context.function(type_hash) {
-                    match unsafe {
-                        //Unsafe block required to avoid cloning index if it isn't needed.
-                        (self as *const Self as *mut Self)
-                            .as_mut()
-                            .unwrap_unchecked()
-                    }
-                    .call_field_fn_prechecked(target, handler, ())?
+                let index = index.clone();
+                return Ok(match self
+                    .call_field_fn(Protocol::GET, target, index.hash(), ())?
                     {
                         CallResult::Ok(()) => return Ok(CallResult::Ok(self.stack.pop()?)),
-                        CallResult::Unsupported(target) => CallResult::Unsupported(target),
-                    }
-                } else {
-                    let index = index.clone();
-                    match self.call_instance_fn(target, Protocol::INDEX_GET, (index,))? {
-                        CallResult::Ok(()) => CallResult::Ok(self.stack.pop()?),
-                        CallResult::Unsupported(target) => CallResult::Unsupported(target),
-                    }
-                });
+                        CallResult::Unsupported(target) => match self.call_instance_fn(target, Protocol::INDEX_GET, (index,))? {
+                            CallResult::Ok(()) => CallResult::Ok(self.stack.pop()?),
+                            CallResult::Unsupported(target) => CallResult::Unsupported(target),
+                        },
+                    });
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/rune/src/runtime/vm.rs
```diff
@@ -984,14 +1022,33 @@ impl Vm {
...
-
-                match self.call_field_fn(Protocol::SET, target, hash, (value,))? {
-                    CallResult::Ok(()) => {
-                        self.stack.pop()?;
-                        CallResult::Ok(())
+                let type_hash =
+                    Hash::field_fn(Protocol::SET, target.type_hash()?, hash.into_type_hash());
+                return Ok(if let Some(handler) = self.context.function(type_hash) {
+                    match unsafe {
+                        //Unsafe block required to avoid cloning index if it isn't needed.
```
udoprog, line 1029: Rework this to get rid of `unsafe` even if it requires cloning some things. Coercing an `&mut self` while borrowing the handler is not appropriate and might result in non-local unsoundness if the implementation was ever changed to do something differently. This should also be using `Vm::call_field_fn`.

udoprog (changes requested):
Thanks for the PR!

Apart the use of `unsafe`: I'm also hesitant to have two protocols bleed together like this. What prevents you from implementing both of them separately on the type you want both operations to work with?

crates/rune/src/runtime/vm.rs
LoopyAshy: that is fine, however the reason I used an alternative implementation to call_field_fn is due to fact call_field_fn did extra work which seemed wasteful, but I am happy to change it and rerun the benchmarks.

regarding having a type having both `GET` and `INDEX_GET` does not necessarily seem like a issue overall, however if you would prefer it could be possibly to have types opt into the fallback functionality?

LoopyAshy: I have separated the Protocols from INDEX_GET and INDEX_SET to FALLBACK_GET and FALLBACK_SET.

PR #449, opened 2023-03-15
Add `contains` instance method for strings

blurgyy:
I currently use a dirty hack to achieve the functionality of `contains` method:

```rust
fn contains(s, sub) {
  s.len() != s.replace(sub, "").len()
}
```

It would be nice if Rune just supports `.contains` IMHO.


udoprog:
Thank you!

Issue #469, opened 2023-03-29
std::future::join enhancement


Rycieos:
I have an `async` function that `join`s a few Futures together. The issue is that these Futures return a value, and `join` makes it hard to retrieve these values.

The included example works great, if you are calling the same function multiple times and don't care about which result is from which call. But this is my use case:

```rune
use std::future;

pub async fn function() {
    let results = future::join([
        { async || ("foo", foo().await) }(),
        { async || ("bar", bar().await) }(),
    ]).await;

    let data = #{};
    for result in results {
        data[result.0] = result.1;
    }

    `[${data.foo}@${data.bar}]`
}
```

This is pretty verbose. I would like an interface closer to this:
```rune
use std::future;

pub async fn function() {
    let data = future::join_to_object({
        foo: foo,
        bar: bar,
    }).await;

    `[${data.foo}@${data.bar}]`
}
```

udoprog:
Here's a slightly less verbose version:

```rust
pub async fn function() {
    let [foo, bar]= future::join([foo(), bar()]).await;
    let data = { foo, bar };
    `[${data.foo}@${data.bar}]`
}
```

Other than that, I'm pretty open to adding more functions for joining futures!

Rycieos:
I forgot to mention that I already tried that:

```rune
    let [foo, bar] = future::join([
        foo(),
        bar(),
    ]).await;
```

> Error: cannot read, value is exclusively accessed (at inst 4)

Same error if first saving as a tuple and later expanding:

```rune
    let result = future::join([
        foo(),
        bar(),
    ]).await;

    let foo = foo.0
```

Maybe it's my fault, since these functions return Rust `&str`s? Is there a different type I should be returning to be "async safe"?

udoprog:
Is there any chance you can provide a working reproduction? A PR to this repo where you add it to examples would be very helpful.

Rycieos:
As resolved in the linked PR, clearly the more complicated version that I needed was caused by my mistake. So your short version does work, but I would like to see a shorter version still.

The problem with the `let [foo, bar] = future::join([foo(), bar()]).await;` form is two fold:
1. The compiler issues a warning `let binding might panic`, which is only fixed by surrounding it in an `if` block (and then the variables are local to that block and can't be used outside of it).
2. The definition of the variable and the function who's result will be assigned to it are separated. This makes keeping them in sync more work.

Hence my suggestion:
```rust
let data = future::join_to_object(#{
    foo: foo(),
[...]

PR #471, opened 2023-03-29
Add example to show instance fn with inner error

Rycieos:
Note that the instance functions returning a `clone()`ed version as a String instead is no different.

As requested in https://github.com/rune-rs/rune/issues/469#issuecomment-1488876246

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -10,7 +10,7 @@ full = ["rune-modules/full"]
 default = ["full"]
 
 [dependencies]
-tokio = { version = "1.26.0", features = ["macros"] }
+tokio = { version = "1.26.0", features = ["macros", "rt-multi-thread"] }
 
 rune = { path = "../crates/rune" }
 rune-modules = { path = "../crates/rune-modules" }
--- a/examples/examples/async_join_externals.rs
+++ b/examples/examples/async_join_externals.rs
@@ -1,0 +1,86 @@
+use std::sync::Arc;
+
+use rune::runtime::Vm;
+use rune::termcolor::{ColorChoice, StandardStream};
+use rune::{Any, ContextError, Diagnostics, FromValue, Module};
+
+#[derive(Debug)]
+struct Inner {
+    foo: String,
+    bar: String,
+}
+
+#[derive(Debug, Any)]
+struct External {
+    map: Inner,
+}
+
+impl External {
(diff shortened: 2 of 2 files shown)


udoprog:
Makes sense, both futures capture `&mut self`. So you can't run them concurrently, same would happen with regular futures (only they'd be checked at compile time):
```rust
struct External;

impl External {
    async fn foo(&mut self) -> &str {
        todo!()
    }

    async fn bar(&mut self) -> &str {
        todo!()
    }
}

#[tokio::main]
async fn main() {
    let mut e = External;
    tokio::join!(e.foo(), e.bar());
}
```
[Playground Link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e37ff596bf85f33e32f9eda61a7cc53d).

If you rework them to somehow take `&self` and return a future, this should not longer be an issue.

Thanks for the example!

Rycieos:
Ah, that makes complete sense. An obvious solution is to use a `RwLock` inside the `External` so that the instance functions are no longer `mut`:

```rust
use std::collections::HashMap;
use std::sync::RwLock;

struct External {
    guard: RwLock<HashMap<&'static str, String>>,
}

impl External {
    async fn foo(&self) -> String {
        let map = self.guard.read().unwrap();
        if map.contains_key("foo") {
            return map.get("foo").unwrap().clone();
        }
        drop(map);

        let value = "FOO".to_string();
        
        let mut map = self.guard.write().unwrap();
        map.insert("foo", value.clone());
        
        value
    }

    async fn bar(&self) -> String {
        let map = self.guard.read().unwrap();
        if map.contains_key("bar") {
            return map.get("bar").unwrap().clone();
        }
        drop(map);

        let value = "BAR".to_string();
        
        let mut map = self.guard.write().unwrap();
        map.insert("bar", value.clone());
        
        value
    }
}

#[tokio::main]
async fn main() {
    let e = External {
        guard: RwLock::new(HashMap::new()),
    };
    
    let results = tokio::join!(e.foo(), e.bar(), e.foo());
    println!("{:?}", results);
}
```
> `("FOO", "BAR", "FOO")`

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0999c094bfc067fcdd7aad30bcc7fe51)

And that works perfectly with Rune code. Thanks for the help!

PR #482, opened 2023-04-06
Language server changes

udoprog:
This pull request significantly improves handling of multiple sources in the language server and adds workspace support.

A workspace is triggered if a `Rune.toml` file is found at the root of the workspace. The sources associated with a workspace are always collected and processed by the language server.

On top of this, any sources you open up will be handled as normal entrypoints unless they are part of a workspace.

This means we can finally deal with non-trivial project topologies and most improtant modules. With the caveat that these must have an associated `Rune.toml` file which points out the entrypoint.

Workspace configuration also comes with its own diagnostics:

image

Hovering over the `rune-languageserver` bottom left option reveals a few controls:

image

Stopping the server prevents it from running at all (during this session) in case there are any issues.


udoprog:
This is all part of [rune-vscode `0.12.7`](https://marketplace.visualstudio.com/items?itemName=udoprog.rune-vscode) for anyone who wants to take it for a spin. It should automatically download the latest nightly languageserver which comes with all the necessary improvements.

PR #490, opened 2023-04-27
Add completion support to the language server

tgolsson:
Very rudimentary start for this I've been hacking at. It's *very* basic; but adds the basic infrastructure that can be improved.

The lookup starts by doing a "symbol + terminal" lookup. Likely this could be split into two checks; but I've opted to do one right now. This happens by searching back from the current cursor until a "terminal". So for example, given:
```rune
pub fn main() {
    let a =[1, 2, 3];
    fo|
}
```
We'd capture `;\n    fo`. The first character is what we use to check if we should lookup instance functions; and then we trim and use the rest for matching.

Then, we go through the unit and find all matches for the symbol - it doesn't seem like the debug info knows whether a rune fn is an instance function or not right now, so it'll return all.

Then, we lookup either all instance fns or all loose fns. It does seem like a lot of member functions also show up as loose fns functions, maybe due to receiver syntax.

In each case, we'll lookup as much possible regarding the item (i.e, docs, args, etc) and pass it along. There's a Resolve API to defer this a bit, but I'm not sure how to use it effectively etc. So I've opted to just return everything immediately.

Things I've noticed but haven't fixed:

* If the compile fails at all *when opening* the server there'll be no Rune-based completion until it succeeds again.
  * This means any custom Rust code will cause the LSP server to fail completion (though that's already the case with diagnostics)
* There's less info available for rune functions
* If you try to complete something common like `iter` you'll have... lots of almost identical matches.
* Completion of local variables or instance fields
* Actually, anything *except* functions aren't handled.
* I've jumped around in almost all .rn files in the repo and haven't had any issues; but... testing this is a bit iffy.
* I don't think I can see imports right now, so all function uses are absolute. If we could see this, we could prefer those items and only look at "non-imported" as secondary sources (+ auto-import!).

### DIFF
--- a/crates/rune/Cargo.toml
+++ b/crates/rune/Cargo.toml
@@ -20,7 +20,7 @@ bench = []
 workspace = ["toml", "semver", "relative-path", "serde-hashkey"]
 doc = ["rust-embed", "handlebars", "pulldown-cmark", "syntect"]
 cli = ["doc", "bincode", "atty", "tracing-subscriber", "anyhow/std", "clap", "webbrowser", "capture-io", "disable-io", "languageserver"]
-languageserver = ["lsp", "ropey", "percent-encoding", "url", "serde_json", "tokio", "workspace"]
+languageserver = ["lsp", "ropey", "percent-encoding", "url", "serde_json", "tokio", "workspace", "doc"]
 capture-io = ["parking_lot"]
 disable-io = []
 
--- a/crates/rune/src/build.rs
+++ b/crates/rune/src/build.rs
@@ -1,7 +1,7 @@
 use crate::ast::Span;
 use crate::compile;
 use crate::compile::{
-    CompileVisitor, FileSourceLoader, NoopCompileVisitor, Options, Pool, SourceLoader,
+    CompileVisitor, FileSourceLoader, Options, Pool, SourceLoader,
 };
 use crate::runtime::Unit;
 use crate::{Context, Diagnostics, SourceId, Sources};
@@ -63,7 +63,7 @@ pub fn prepare(sources: &mut Sources) -> Build<'_> {
         context: None,
         diagnostics: None,
         options: None,
-        visitor: None,
+        visitors: Vec::new(),
         source_loader: None,
     }
 }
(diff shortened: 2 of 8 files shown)


crates/rune/Cargo.toml
```diff
@@ -20,7 +20,7 @@ bench = []
 workspace = ["toml", "semver", "relative-path", "serde-hashkey"]
 doc = ["rust-embed", "handlebars", "pulldown-cmark", "syntect"]
 cli = ["doc", "bincode", "atty", "tracing-subscriber", "anyhow/std", "clap", "webbrowser", "capture-io", "disable-io", "languageserver"]
-languageserver = ["lsp", "ropey", "percent-encoding", "url", "serde_json", "tokio", "workspace"]
+languageserver = ["lsp", "ropey", "percent-encoding", "url", "serde_json", "tokio", "workspace", "doc"]
```
tgolsson, line 23: This is required to use `crate::doc` with the doc visitor.

crates/rune/src/build.rs
```diff
@@ -74,10 +74,59 @@ pub struct Build<'a> {
     context: Option<&'a Context>,
     diagnostics: Option<&'a mut Diagnostics>,
     options: Option<&'a Options>,
-    visitor: Option<&'a mut dyn compile::CompileVisitor>,
+    visitors: Vec<&'a mut dyn compile::CompileVisitor>,
     source_loader: Option<&'a mut dyn SourceLoader>,
 }
 
+/// Wraps a collection of CompileVisitor
+struct CompileVisitorGroup<'a> {
```
tgolsson, line 82: This was required to avoid a `&[& ...]` situation in every API currently taking CompileVisitor.

crates/rune/src/doc/visitor.rs
```diff
@@ -4,6 +4,7 @@ use crate::compile::{
 };
 use crate::hash::Hash;
 
+#[derive(Clone)] // TODO(TSol): this is a bit hackish
```
tgolsson, line 7: This was necessary for the docs lookup; it didn't seem to work right when putting unit and the associated docs on the workspace. Didn't investigate too much.

crates/rune/src/languageserver/state.rs
```diff
@@ -354,24 +563,33 @@ impl<'a> State<'a> {
...
         mut build: Build,
         built: Option<&mut HashSet<Url>>,
-    ) -> (crate::Diagnostics, Build, Visitor) {
+    ) -> (
+        crate::Diagnostics,
+        Build,
+        Visitor,
+        crate::doc::Visitor,
+        Result<Unit, BuildError>,
+    ) {
```
tgolsson, line 572: This probably should be a type now...

crates/rune/src/languageserver/state.rs
```diff
@@ -278,6 +480,12 @@ impl<'a> State<'a> {
 
                 source.index = value;
                 source.build_sources = Some(sources.clone());
+
+                if let Ok(unit) = unit.as_ref().map(|v| v.clone()) {
+                    source.unit = Some(unit.clone());
+                }
+
+                source.docs = Some(doc_visitor.clone());
```
tgolsson, line 488: This is definitely wrong; but there seems to be some missing "level" in between here. (I.e, what's the build target mapping on a workspace?)

crates/rune/src/languageserver/state.rs
```diff
@@ -191,6 +193,205 @@ impl<'a> State<'a> {
...
+        position: lsp::Position,
+    ) -> Option<Vec<lsp::CompletionItem>> {
+        let sources = &self.workspace.sources;
+        tracing::info!("sources: {:?}", sources.get(uri).is_some());
+        let workspace_source = sources.get(uri)?;
+
+        let offset = workspace_source.lsp_position_to_offset(position);
+
+        let (mut symbol, start) = workspace_source.looking_back(offset)?;
+        tracing::info!("symbol : {:?}, start: {:}", symbol, start);
```
tgolsson, line 209: Note to self: remove!

crates/rune/src/languageserver/state.rs
```diff
@@ -191,6 +193,205 @@ impl<'a> State<'a> {
...
     }
 
+    /// Find definition at the given uri and LSP position.
+    pub async fn complete(
+        &self,
+        uri: &Url,
+        position: lsp::Position,
+    ) -> Option<Vec<lsp::CompletionItem>> {
+        let sources = &self.workspace.sources;
+        tracing::info!("sources: {:?}", sources.get(uri).is_some());
```
tgolsson, line 203: Note to self: remove!

crates/rune/src/languageserver/state.rs
```diff
@@ -191,6 +193,205 @@ impl<'a> State<'a> {
...
+        let (mut symbol, start) = workspace_source.looking_back(offset)?;
+        tracing::info!("symbol : {:?}, start: {:}", symbol, start);
+
+        if symbol.is_empty() {
+            return None;
+        }
+
+        let mut results = vec![];
+
+        let can_use_instance_fn: &[_] = &['.'];
```
tgolsson, line 217: Should probably just check `== '.'` below. Though this is a bit clearer documentation-wise.

crates/rune/src/languageserver/state.rs
```diff
@@ -191,6 +193,205 @@ impl<'a> State<'a> {
...
+        let first_char = symbol.remove(0);
+        let symbol = symbol.trim();
+        let untrimmed_length = symbol.len();
+
+        if let Some(unit) = workspace_source.unit.as_ref() {
```
tgolsson, line 222: Should probs move a bunch of this to a new struct... future work maybe?

tgolsson: It gets a bit iffy, whatever takes over this code ("CompletionState?") would still need unit/context...

crates/rune/Cargo.toml
udoprog: 👍

udoprog (approved):
Pretty happy with the state of it as-is. Merge whenever you feel like and we'll clean it up as we get along.

You're starting to do some "analysis on incomplete sources" which is what I eventually want #383 for so that we can unify the two. Since we don't have lossless parsing (yet!) we'll deal with it as it comes along.

Thank you!

crates/rune/src/languageserver/state.rs
tgolsson: Having to clone out the docs to multiple sources feels wrong out to me.

udoprog:
Thank you!

PR #492, opened 2023-05-01
Add `rune fmt` subcommand 

tgolsson:
So, this turned out to be massive. Not really anything complex, but yeah.

There's three components:

* The `Printer` struct responsible for laying out AST nodes
* The `IndentWriter` responsible for building lines and indenting them
* The `SpanInjectionWriter` which dynamically reinjects non-ast nodes into the generated document.

The printer just walks the input AST tree, writing all the data to a buffer. I've not done any efforts to optimize this in any way; I'm sure some macros or generics could clean it up a lot. But... complexity when iterating. And I think that'd fall over again if we want something more heuristics-based.

The second one is the actual writer which dynamically builds lines, indenting as specified. 

The he last one is built on the idea that we're traversing two documents in parallel; one being fed in from the printer and the other ones queued up at the start. Then just print in order. 

A few design decisions: 
* I preferred using full destructuring (instead of `..` or `.`) to avoid breaking changes in the future where the formatter doesn't get updated when new fields are addd
* I opted to not do traits, as I think that'll make heuristics-based formatting (e.g. breaking chains over multiple lines) much harder.
* I've tried to do some dumb heuristics for line-breaking structs etc, instead of trying to understand what the user wanted.

Review guide: I'd do (*everything*) but the `Printer` file, and then try to trace some expressions through the formatter. I'm not sure if reviewing it line-by-line is even possible. :D


udoprog (commented):
Looks really good!

I'm on my phone right now so I'll pull it onto my desktop tomorrow.

It looks like the command is `rune format` (unless I'm missing it on this screen). I'd like it to be `rune fmt` to be in line with `cargo fmt`.

If you could add a `rune fmt --check` as well and add it to CI that would be brilliant.

See other comments. I promise a more closer look tomorrow but I don't really see any technical issues. Well done!

tgolsson:
CI check done :-) 

image

crates/rune/src/fmt/printer.rs
```diff
@@ -0,0 +1,2020 @@
+//! The `Printer` trait and implementations.
+
+use std::io::Write;
+
+use crate::ast::{
```
udoprog, line 5: I *usually* qualify types from here using `ast::AngleBracketed`. But hey, this works too 😆

udoprog (approved):
Only nits I had I just went ahead and fixed. The printer must've been a fun effort, but it's very easy to follow what it's doing. Very good job!

The only thing I'd like to change code structure wise is move the tests into their own modules. Especially with big files it's kinda hard to spot the tests otherwise. But that is not a blocker. If you don't want to change it here I'll do it later.

Thank you again! Merge whenever you feel ready.

.github/workflows/ci.yml
```diff
@@ -56,6 +56,13 @@ jobs:
         components: rustfmt
     - run: cargo fmt --all -- --check
 
+  runefmt:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@stable
+    - run: shopt -s globstar ; cargo run --bin rune -- fmt --check **/*.rn
```
udoprog, line 64: I'd like to encapsulate all the examples in the scripts folder in the workspace somehow, that way this could be called without globbing. But this obviously works for now!

crates/rune/src/fmt/comments.rs
```diff
@@ -106,12 +105,11 @@ fn parse_line_comment(chars: &mut Peekable<impl Iterator<Item = (usize, char)>>)
...
 }
 
-fn parse_block_comment(chars: &mut Peekable<impl Iterator<Item = (usize, char)>>) -> Option<usize> {
+fn parse_block_comment(chars: &mut CharIndices<'_>) -> Option<usize> {
     while let Some((_, c)) = chars.next() {
         if c == '*' {
-            if let Some((_, '/')) = chars.peek() {
-                let (offset, _) = chars.next()?;
-                return Some(offset);
+            if let Some((_, '/')) = chars.clone().next() {
```
udoprog, line 111: This pattern is actually kinda interesting: For character iterators peekable is pretty much a waste. The char iterator itself is three pointers big - a slice iterator (two pointers) and the index (one usize). Cloning it to peek is actually fine.

udoprog:
I've merged master in and there was a minor conflict in Cargo.toml. Whenever you're ready to merge this feel free to squash the history.

crates/rune/src/fmt/printer.rs
tgolsson: It became a bit too noisy :-)

.github/workflows/ci.yml
tgolsson: I think it'd be nice with a `-R` flag or something on the script; just felt like it'd need some more changes to the CLI entrypoint.

tgolsson:
> Only nits I had I just went ahead and fixed. The printer must've been a fun effort, but it's very easy to follow what it's doing. Very good job!

Thank you! Yeah the printer was a fun thing; it was *very* to bash out actually. A bit hard to track where things like extra spaces or newlines come from though post-fact... 

> The only thing I'd like to change code structure wise is move the tests into their own modules. Especially with big files it's kinda hard to spot the tests otherwise. But that is not a blocker. If you don't want to change it here I'll do it later.

I'll move them out, sure! I've got some refactoring ideas to make it a bit easier to test in the future. By doing functions like `loose layout_function(ast, formatting_constraints) -> Result<String>` it's easier to test individual layout functionality, and more importantly do adaptive handling of line-lengths. That's a bit closer to how rustfmt does it too.

tgolsson:
Just saw you'd moved the tests. So all good from your side? :-) Then I think we can go ahead and merge. I've got nothing more to do I think.

.github/workflows/ci.yml
udoprog: Just remembered that recursive does exist with `--recursive`.

PR #520, opened 2023-05-22
fix languageserver for neovim by not borrowing method in IncomingMessage

ModProg:
- add Context::iter_functions to languageserver feature
- fix languageserver for neovim by not borrowing method in IncomingMessage

### DIFF
--- a/crates/rune/src/compile/context.rs
+++ b/crates/rune/src/compile/context.rs
@@ -237,7 +237,7 @@ impl Context {
     }
 
     /// Iterate over all available functions in the [Context].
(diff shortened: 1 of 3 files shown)


udoprog:
Huh, which `method` is being used? I thought all of them could be decoded verbatim 😄

ModProg:
> Huh, which `method` is being used? I thought all of them could be decoded verbatim smile

I guess they should be, but it sends the open event as `textDocument\/didOpen` which is apparently valid json.

ModProg:
probably whatever json library they use is a bit eager to escape.

PR #528, opened 2023-05-26
Attribute Macros on Items

ModProg:
- [x] tests
- [x] `rune::attribute_macro`

### DIFF
--- a/crates/rune/src/ast/attribute.rs
+++ b/crates/rune/src/ast/attribute.rs
@@ -60,6 +60,13 @@ pub struct Attribute {
     /// The `]` character
     pub close: T![']'],
 }
+impl Attribute {
+    pub(crate) fn stream_span(&self) -> Span {
+        self.input
+            .option_span()
+            .unwrap_or_else(|| self.close.span.head())
+    }
+}
 
 impl Parse for Attribute {
     fn parse(p: &mut Parser<'_>) -> Result<Self> {
--- a/crates/rune/src/ast/attribute.rs
+++ b/crates/rune/src/ast/attribute.rs
@@ -61,7 +61,7 @@ pub struct Attribute {
     pub close: T![']'],
 }
 impl Attribute {
-    pub(crate) fn stream_span(&self) -> Span {
+    pub(crate) fn input_span(&self) -> Span {
         self.input
             .option_span()
             .unwrap_or_else(|| self.close.span.head())
--- a/crates/rune/src/compile/context.rs
+++ b/crates/rune/src/compile/context.rs
@@ -634,7 +634,7 @@ impl Context {
         self.install_meta(ContextMeta {
             hash,
             item: Some(item),
-            kind: meta::Kind::Macro,
+            kind: meta::Kind::AttributeMacro,
             #[cfg(feature = "doc")]
             docs: m.docs.clone(),
         })?;
--- a/crates/rune/src/compile/context.rs
+++ b/crates/rune/src/compile/context.rs
@@ -9,12 +9,12 @@ use crate::compile::meta;
 use crate::compile::Docs;
 use crate::compile::{ComponentRef, ContextError, IntoComponent, Item, ItemBuf, MetaInfo, Names};
 use crate::module::{
-    Fields, Function, InternalEnum, Module, ModuleAssociated, ModuleConstant, ModuleFunction,
-    ModuleMacro, ModuleType, TypeSpecification, UnitType,
+    Fields, Function, InternalEnum, Module, ModuleAssociated, ModuleAttributeMacro, ModuleConstant,
+    ModuleFunction, ModuleMacro, ModuleType, TypeSpecification, UnitType,
 };
 use crate::runtime::{
-    ConstValue, FunctionHandler, MacroHandler, Protocol, RuntimeContext, StaticType, TypeCheck,
-    TypeInfo, TypeOf, VariantRtti,
+    AttributeMacroHandler, ConstValue, FunctionHandler, MacroHandler, Protocol, RuntimeContext,
+    StaticType, TypeCheck, TypeInfo, TypeOf, VariantRtti,
 };
 use crate::Hash;
 
@@ -92,6 +92,8 @@ pub struct Context {
     associated: HashMap<Hash, Vec<Hash>>,
     /// Registered native macro handlers.
     macros: HashMap<Hash, Arc<MacroHandler>>,
+    /// Registered native attribute macro handlers.
+    attribute_macros: HashMap<Hash, Arc<AttributeMacroHandler>>,
     /// Registered types.
     types: HashMap<Hash, ContextType>,
     /// Registered internal enums.
@@ -217,6 +219,10 @@ impl Context {
             self.install_macro(module, m)?;
         }
 
+        for m in &module.attribute_macros {
+            self.install_attribute_macro(module, m)?;
+        }
+
         for m in &module.constants {
             self.install_constant(module, m)?;
         }
@@ -329,6 +335,11 @@ impl Context {
     /// Lookup the given macro handler.
     pub(crate) fn lookup_macro(&self, hash: Hash) -> Option<&Arc<MacroHandler>> {
         self.macros.get(&hash)
+    }
+
+    /// Lookup the given attribute macro handler.
+    pub(crate) fn lookup_attribute_macro(&self, hash: Hash) -> Option<&Arc<AttributeMacroHandler>> {
+        self.attribute_macros.get(&hash)
     }
 
     /// Look up the type check implementation for the specified type hash.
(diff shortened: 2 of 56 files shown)


udoprog:
It looks good to me, I've merged your changes onto main after merging #526 since they both touch on the same stuff.

crates/rune/src/compile/context.rs
```diff
@@ -610,6 +621,27 @@ impl Context {
...
+        m: &ModuleAttributeMacro,
+    ) -> Result<(), ContextError> {
+        let item = module.item.join(&m.item);
+        let hash = Hash::type_hash(&item);
+        self.attribute_macros.insert(hash, m.handler.clone());
+
+        self.install_meta(ContextMeta {
+            hash,
+            item: Some(item),
+            kind: meta::Kind::Macro,
```
udoprog, line 637: This should use a new kind, `meta::Kind::Macro` is expected to be looked up as function-style macros, like `hello!(..)`.

We might also need to do some conflict resolution in case an attribute is named the same as a function-style macro. I.e. in [call contexts](https://github.com/rune-rs/rune/blob/e5b7a3a1095768b26fa05ec4e2857aba78a019f7/crates/rune/src/compile/v1/assemble.rs#L2773) we should prefer to use the latter since you can't *call* an attribute.

You should add a test case to see the issue:
* Register an attribute and a regular macro no the same path, both called `conflict`.
* Try to build an expression like `conflict!()`.

crates/rune/src/macros/macro_compiler.rs
```diff
@@ -74,4 +76,67 @@ impl MacroCompiler<'_> {
...
+                    span,
+                    CompileErrorKind::MissingMacro {
+                        item: self.query.pool.item(named.item).to_owned(),
+                    },
+                ));
+            }
+        };
+
+        // TODO ATTRMCRO this should maybe remove the leading `=` or surrounding delimiter
+        let input_stream = &attribute.input;
```
udoprog, line 122: Hm, yeah. This would require some light parsing which is annoying. In Rust the AST is made up of [groups](https://doc.rust-lang.org/stable/proc_macro/enum.TokenTree.html#variant.Group) which are easy enough to strip.

Did you ever try what happens with attribute macros using `=` in Rust? I haven't checked that out yet.

crates/rune/src/module/function_meta.rs
```diff
@@ -111,6 +112,31 @@ impl FunctionMacroData {
...
+/// Runtime data for an attribute macro.
+#[derive(Clone)]
+pub struct AttributeMacroData {
+    pub(crate) item: ItemBuf,
+    pub(crate) handler: Arc<AttributeMacroHandler>,
+}
+
+impl AttributeMacroData {
+    #[inline]
+    pub(crate) fn new<F, N>(name: N, f: F) -> Self
```
udoprog, line 124: This is never used (Clippy warning), Using it would require extending the `rune::macro_` attribute since that's where it's used. It's also something we should do since it allows for registering the attribute macro using `Module::macro_meta` (the new cool way to do it).

crates/rune/src/ast/attribute.rs
```diff
@@ -60,6 +60,13 @@ pub struct Attribute {
     /// The `]` character
     pub close: T![']'],
 }
+impl Attribute {
+    pub(crate) fn stream_span(&self) -> Span {
```
udoprog, line 64: Maybe this should be called `input_span`?

udoprog (commented):
Looks good to me. Some missing pieces and possible meta conflict, but all the infrastructure to run the macros seem to be there. Good job!

crates/rune/src/macros/macro_compiler.rs
udoprog: So it seems like key-value attributes are not supported in Rust:

image

udoprog: I think this might be an instance where it's worth deviating from Rust. The restrictions imposed by key-value attribute macros seems to have been [a historical accident](https://github.com/rust-lang/rust/issues/34981). Us having to reparse macro input just to clean it up seems silly.

If so, it seems natural to parse on the input as you're already providing it since it leads to unambiguous parsing requirements across styles. To facilitate parsing it would be a good idea to implement a helper function for parsing groups, [which implements this balancing logic](https://github.com/rune-rs/rune/blob/main/crates/rune/src/ast/macro_call.rs#L74).

What do you think?

ModProg: makes sense

crates/rune/src/ast/attribute.rs
ModProg: maybe, the problem is, it'll end up in a MacroContext where it will be the "stream_span" maybe we should add a new AttributeMacroContext, but that would be a problem with the requirnment of passing macro cobtext around everywhere. AFAICT this is for error spans, but one could maybe also just return an error that says "Ihave no span yet give me the surrounding one" and the parrent function would apply the correct span instead of passing through the macrocontext to provide the span.

udoprog: I was just thinking that it reflects the name of the field. At the very least they should be named the same 😄.

ModProg: I mean one could argue that `input` is the correct name for function macros as well

ModProg: but not sure what you think of that breaking change.

crates/rune/src/macros/macro_compiler.rs
ModProg: I could also imagine having an AST element for a delimited group that one could use to just call parse with.

ModProg:
well this broke builtin attributes such as `#[test]`.

Would it be possible to actually impllement them as attribute macros? Or should we somehow skip them and keep them arround though I'm not sure how that would work best.

udoprog:
Skip over attributes you can't process until the end, then error if any unprocessed attributes remain?

ModProg:
> Skip over attributes you can't process until the end, then error if any unprocessed attributes remain?

So I don't error on unknown attributes, but just ignore them so that the item handler would error later.

udoprog:
Hm, I'm not entirely following. What is the problem right now? I'll also look over the PR later, so maybe I'll better understand what's going on then :sweat_smile:

ModProg:
So I'm really not sure how to best implement this, e.g. in this:

```rs
#[some_macro]
#[test]
#[another_macro]
fn foobar() {}
```

This should expand `some_macro` and `another_macro` and remove both, but keep `test` so this function is a test.

Currently I use `remove_first_attribute` which will pop the first attribute (some_macro) and resolve + expand it, afterwards it would pop `test`, but cannot resolve it and needs to somehow preserve it in a way where the next call to `remove_first_attribute` would remove `another_macro`.

In my current setup this is actually reasonably possible, that is, because I haven't implemented lenient import resolution i.e.:
```
#[attribute] fn smth() {}
use ::module::attribute;
```
is currently an error.

To support the order agnostic usage, I would need to:
1. expand macros until one is unknown
2. enqueue annotated item and continue expanding other items
3. when there are no expandable macros and no imports left, skip the first macro and try to expand again, repeating until all items with unresolved attributes are finished.

It is not possible to skip the top attribute if unknown and expand the ones below, because an unknown attribute might remove said attributes which would have been erroneously expanded.

From that, I think it would be best to filter built-in attributes directly, (maybe directly in `remove_first_attribute`) or if possible implement them as actual macros.

I can as an MVP support attributes that are resolved in order, i.e., only if the import was before the attribute, and just skip unknown attributes and leave dealing with them to `item_fn`.

udoprog:
> Currently I use `remove_first_attribute` which will pop the first attribute (some_macro) and resolve + expand it, afterwards it would pop `test`, but cannot resolve it and needs to somehow preserve it in a way where the next call to `remove_first_attribute` would remove `another_macro`.

This shouldn't be an issue if you keep any attributes which you can't expand and hand it to downstream steps, which will either process them or error. But due note that `#[some_macro]` would be fed as input `#[test] #[another_macro] fn foobar() {}`, so it's up to that macro handler what exactly to produce. But whatever it produces would be recursively expanded - if more attributes are found.

Let's say `#[test] #[another_macro] fn foobar() {}` is produced next, then `#[test]` is ignored, and simply prepended as attributes to whatever `#[another_macro] fn foobar() {}` produces.

Does this make sense?

> lenient import resolution

The type of import resolution currently implemented simply means that macros are "evaluated last". So what you need to do is look over available items once - and if they contain attributes that could be expanded by a macro put them in the queue to be expanded after all other items. That way uses (which are not production of macros) are always expanded first. And the rest of evaluation is order-dependent from top-to bottom as they are added to the queue.

So to determine whether something should be deferred or not, I think some kind of first pass needs to be performed to test if there are attributes that need to be expanded.

ModProg:
that does make it more easily achievable. I'll look into it!

udoprog:
I'm nursing a cold right now, will get into it when I'm 100% again!

crates/rune/src/module/module.rs
```diff
@@ -571,17 +577,37 @@ impl Module {
...
                 });
+                &mut self.macros.last_mut().unwrap().docs
             }
-        }
+            MacroMetaKind::Attribute(data) => {
+                let hash = Hash::type_hash(&data.item);
 
-        let m = self.macros.last_mut().unwrap();
-        Ok(ItemMut { docs: &mut m.docs })
+                if !self.names.insert(Name::Macro(hash)) {
```
udoprog, line 585: Isn't this supposed to be `Name::AttributeMacro`?

udoprog (commented):
Thank you, this is a great start. Only the `Name` issue (I think).

crates/rune/src/module/module.rs
ModProg: ups, musst have missed that one

udoprog (approved):

udoprog:
Thank you!

PR #531, opened 2023-05-27
Fix derive macros outside `rune`

### LINKED ISSUE #530: Make derive macros work outside of `rune`
Make derive macros work outside rune such as derive(Parse). To do this one can use either https://docs.rs/proc-macro-crate/latest/proc_macro_crate/ directly or a more complicated approach like here: https://github.com/khonsulabs/bonsaidb/blob/448db4d32ea065aa39b1c4731c0e5b9d5ffce83b/crates/bonsaidb-macros/src/lib.rs#L35-L61 to get more control about it.

ModProg:
fixes #530


udoprog:
Honestly, not worth the extra complexity (esp. w.r.t. manifest parsing) and dependency: All public macros are re-exported from `rune`, and all you need to make them work internally is `use crate as rune` (you can see this in some places).

If one of the public macros do not use `rune::` as a default prefix (which I think most of them do), it should be fixed.

PR #532, opened 2023-05-27
Support generics in more derive macros

ModProg:
- use `rune` instead of `crate` for paths in derive macros
- support generics in `derive(Parse, ToTokens, Spanned)`


udoprog:
Cool, very nice!

PR #539, opened 2023-06-03
fix and add tests for macrocall whitespace

tgolsson:
Also found another small bug where any whitespace and comments at EOF would disappear; as it would be after the last span we re-emit from the AST. These are now unconditionally re-emitted during commit.


udoprog:
Nice!

PR #544, opened 2023-06-05
add `Protocol::TRY`

ModProg:
- [x] added Protocol::TRY and call it in `?`
- [x] add tests
- [x] bikeshed the name

### DIFF
--- a/crates/rune-core/src/protocol.rs
+++ b/crates/rune-core/src/protocol.rs
@@ -424,8 +424,8 @@ impl Protocol {
         /// it should return `Result<T, Result<(), E>>`
         pub const TRY: Protocol = Protocol {
             name: "try",
-            hash: 0xe44ad51ace6f0cd6,
-            repr: None,
+            hash: 0x5da1a80787003354,
+            repr: Some("value?"),
             doc: ["Allows the `?` operator to apply to values of this type."],
         };
     }
--- a/crates/rune-core/src/protocol.rs
+++ b/crates/rune-core/src/protocol.rs
@@ -414,5 +414,19 @@ impl Protocol {
             repr: None,
             doc: ["Test if the provided argument is a variant."],
         };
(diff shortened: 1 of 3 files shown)


ModProg:
What would the best place to add a test?

Should we keep `TRY` as the name?

I just used the hash calling `.try()` printed, not sure if that is what you
want.

crates/rune-core/src/protocol.rs
```diff
@@ -414,5 +414,19 @@ impl Protocol {
...
+        ///
+        /// Signature: `fn(self) -> Result`.
+        ///
+        /// Note that it uses the `Result` like [`std::ops::Try`] uses
+        /// [`ControlFlow`](std::ops::ControlFlow) i.e., for `Result::<T, E>`
+        /// it should return `Result<T, Result<(), E>>`
+        pub const TRY: Protocol = Protocol {
+            name: "try",
+            hash: 0xe44ad51ace6f0cd6,
+            repr: None,
```
udoprog, line 428: repr: Some("value?"),

udoprog:
> I just used the hash calling .try() printed, not sure if that is what you
want.

Commented elsewhere that they're randomly generated. I'm not exactly sure how you did the above, but if it's at risk of conflicting with something else in the future it's better to avoid!
```sh
cargo run --bin rune -- hash --random
```

udoprog:
> What would the best place to add a test?

Sorry, missed this. In the `rune` crate. I moved the tests in here so that they can make use of `pub(crate)` items and compile faster (linking is exceedingly slow on some platforms).

You can see a bunch of tests in `crates/rune/src/tests`. Just remember to add the module in `tests.rs`.

ModProg:
I'll add a test later, already tested it in my project and it works!

udoprog:
Cheers!

PR #548, opened 2023-06-07
More HIR lowering

udoprog:
This sets the groundwork for performing variable tracking during HIR lowering as well, which will help with future SSA construction.

Context has been moved as a reference into `Query`, which greatly simplified *a lot* of things.


udoprog:
Managed to move capture analysis to the lowering phase, which removeds a long standing issue. Since indexing doesn't have access to metadata (it can't, because metadata visibliity requires indexing to complete!) it can't distinguish between local variables and meta. We've previously "fuzzed" this by saying that anything that looks like a variable during indexing (has a declaration and is a limited path) is captured. However, this can't follow the same behavior as the rest of the program.

Now since it's a part of lowering, variable:ness decisions are made consistently regardless of where it's done.

PR #562, opened 2023-06-12
Tuple and VecTuple incorrectly takes their values

DarkRTA:
This fixes a bug where converting a value to a Rust tuple will incorrectly take ownership of it.


udoprog:
I extended the PR to fix VecTuple at the same time and add a test case.

Thank you again for bringing this to our attention!

PR #576, opened 2023-06-30
Support formatting in lsp

ModProg:


crates/rune/src/languageserver/state.rs
```diff
@@ -250,6 +250,34 @@ impl<'a> State<'a> {
...
     }
 
+    pub(super) fn format(&mut self, uri: &Url) -> Option<TextEdit> {
+        let sources = &mut self.workspace.sources;
+        tracing::trace!(uri = ?uri, uri_exists = sources.get(uri).is_some());
+        let workspace_source = sources.get_mut(uri)?;
+
+        // Not sure how to get source of `workspace_source`
+        let source = crate::source::Source::new(uri, workspace_source.content.to_string());
+        let formatted = crate::fmt::layout_source(&source).ok()?;
```
udoprog, line 260: The signature of layout_source doesn't make much sense forcing you to roundtrip, it could be simpler:
<https://github.com/rune-rs/rune/blob/96686814ad7626f99265856b6fad9c959d98bec3/crates/rune/src/fmt.rs#L22>.

crates/rune/src/languageserver/state.rs
```diff
@@ -250,6 +250,34 @@ impl<'a> State<'a> {
...
+        let formatted = String::from_utf8(formatted).expect("format only produces valid utf8");
```
udoprog, line 261: Err instead of panic. The former is handled by the server.

crates/rune/src/languageserver/state.rs
```diff
@@ -250,6 +250,34 @@ impl<'a> State<'a> {
...
+
+        // Only modify if changed
+        if source.as_str() != formatted {
+            workspace_source.content = Rope::from_str(&formatted);
+            self.rebuild_interest();
+
+            Some(TextEdit::new(
```
udoprog, line 268:
```suggestion
            Some(lsp::TextEdit::new(
```

crates/rune/src/languageserver/state.rs
```diff
@@ -7,7 +7,7 @@ use crate::no_std::collections::HashMap;
 use crate::no_std::prelude::*;
 
 use anyhow::{anyhow, Context as _, Result};
-use lsp::Url;
+use lsp::{TextEdit, Url};
```
udoprog, line 10:
```suggestion
use lsp::Url;
```

crates/rune/src/languageserver/state.rs
udoprog, line 260: Some kind of user-feedback when formatting errors would be nice too. AFAICT, this will no-op.

udoprog (commented):
Thanks! Looks mostly good, some nits and suggestions.

crates/rune/src/languageserver/state.rs
ModProg: if it doesn't fmt that would be a syntax error right? And a syntax error is reported anyway?

e.g. rust-analyzer also doesn't error when fmt fails. (I think only wen trailing whitespace remains)

crates/rune/src/languageserver/state.rs
ModProg: Right, i just expected fmt to check this already meaning that this error can never happen.

crates/rune/src/languageserver/state.rs
ModProg: what exactly do you mean by that?

crates/rune/src/languageserver/state.rs
ModProg: Only thing I see would be to return `String` instead of `Vec<u8>`?

crates/rune/src/languageserver/state.rs
ModProg: Only thing I see would be to return `String` instead of `Vec<u8>`?

crates/rune/src/languageserver/state.rs
udoprog: Yeah I get it. That might be true today, but it's a big invariant to remember and maintain. The outcome of a panic is the same as erroring (language server goes down), but with worse diagnostics.

crates/rune/src/languageserver/state.rs
udoprog: Maybe, then do what you feel is best.

udoprog: I was thinking about changing the argument to be `source: &str`, I don't think the `Source` type is doing anything useful.

Since you didn't add it, I won't force you to do it. Only if you want to and no other issues pop up as a result.

udoprog (approved):

udoprog:
Much appreciated! Should be out in the next nightly.

PR #584, opened 2023-07-19
Support implementing protocols through `#[rune::function]`

ModProg:


crates/rune-macros/src/function.rs
```diff
@@ -208,15 +223,24 @@ impl Function {
...
+                })
+            } else {
+                syn::Expr::Lit(syn::ExprLit {
+                    attrs: Vec::new(),
+                    lit: syn::Lit::Str(match &attrs.path {
+                        Path::None => name_string.clone(),
+                        Path::Rename(last) | Path::Instance(_, last) => {
+                            syn::LitStr::new(&last.ident.to_string(), last.ident.span())
+                        }
+                        Path::Protocol(_) => unreachable!(),
```
udoprog, line 240: I would prefer some repetition over using `unreachable!`. Building a lit expr is not very hard and I'd be worried about exporting the branch during future refactoring.

Alt something like this is also OK imo:
```rust
name = 'out: {
    let lit = match &attrs.path {
        Path::Protocol(protocol) => break 'out syn::Expr::Path(..),
        Path::None => name_string.clone(),
        Path::Rename(last) | Path::Instance(_, last) => {
            syn::LitStr::new(&last.ident.to_string(), last.ident.span())
        }
    };

    syn::Expr::Lit(..)
}
```

crates/rune-macros/src/function.rs
```diff
@@ -36,6 +37,14 @@ impl FunctionAttrs {
                 out.instance = true;
             } else if ident == "keep" {
                 out.keep = true;
+            } else if ident == "protocol" {
+                input.parse::<Token![=]>()?;
+                let protocol: syn::Path = input.parse()?;
+                out.path = Path::Protocol(if protocol.segments.len() == 1 {
+                    parse_quote_spanned!(protocol.span()=> rune::runtime::Protocol::#protocol)
```
udoprog, line 44: :thinking: I do have a preference towards building syn structures manually as much as possible, but I suppose this is fine. I would note that building and manipulating a `syn::Path` by hand is not hard.

udoprog:
Thanks! Please also do add some documentation and tests.

crates/rune-macros/src/function.rs
ModProg: I saw that. Personally, I don't think the type-safety one gets by that really offsets the convenience of `quote!`, but I can change it.

udoprog:
Thank you!

PR #587, opened 2023-07-25
allow to derive protocols with Any

### LINKED ISSUE #583: `derive(Any)`  type protocols
What I want is some way to use e.g. derived functions or trait implementations in general with `derive(Any)`.

```rs
#[derive(Any, Debug, Display)]
#[display(fmt = "name = {value}")]
// STRING_DISPLAY uses the `Display` trait implementation
#[rune(item = ::std::env, protocols(STRING_DISPLAY, DEBUG = Self::debug)]
struct Var {
    name: String,
}

impl Var {
    // This would map debug to display instead of using the rust debug.
    fn debug(&self, buf: &mut String) -> fmt::Result {
        write!(buf, "{self}");
    }
}
```

I have put together something here: https://github.com/rune-rs/rune/commit/bbb389798dc9a5618c4bb596bb4e501865b52837 though I think it could have a nicer interface:
```rs
#[derive(Any, Clone, Display)]
#[rune(item = ::std::env, protocols = [STRING_DISPLAY])]
#[display(fmt = "{}", "value.as_deref().unwrap_or_default()")]
struct Var {
    name: String,
    value: Option<String>,
}
```

ModProg:
fixes #583


ModProg:
This now also includes 0adc1bb161d3f36363cff6f695386a0a2ac36b45, which is related but not equivalent for associated functions.

```rs
#[derive(Any, Clone)]
#[rune(item = ::std::env)]
#[rune(functions(Self::is_set))]
struct Var {
    name: String,
    value: Option<String>,
}

impl Var {
    #[function]
    fn is_set(&self) -> bool {
        self.value.is_some()
    }
}
```

ModProg:
Also, very much WIP, if you agree overall, I'll add the missing traits. And we'll also need documentation and tests.

udoprog:
Thanks!

I'm out now! I'll check it out when I get back next week.

crates/rune-macros/src/context.rs
```diff
@@ -145,7 +203,7 @@ impl Context {
         self.errors.borrow_mut().push(error)
     }
 
-    /// Get a field identifier.
+    /// Get a field identifier.FunctionMetaKind
```
udoprog, line 206: Rogue identifier

crates/rune-macros/src/context.rs
```diff
@@ -111,6 +115,60 @@ pub(crate) struct FieldProtocol {
...
+            "STRING_DEBUG" => quote_spanned! {self.protocol.span()=>
+                module.associated_function(rune::runtime::Protocol::STRING_DEBUG, |this: &Self, buf: &mut String| {
+                    use ::std::fmt::Write as _;
+                    ::std::write!(buf, "{this:?}")
+                })?;
+            },
+            _ => syn::Error::new_spanned(
+                &self.protocol,
+                format!(
+                    "`{}` is not a protocol supported for automatic generation on a type",
```
udoprog, line 150: Hm, what about "Rune protocol `{}` cannot be derived"?

Would be nice to make it clearer where the error message originates from, and for it to be a bit more concise.

crates/rune-macros/src/context.rs
```diff
@@ -111,6 +115,60 @@ pub(crate) struct FieldProtocol {
...
+impl TypeProtocol {
+    pub fn expand(&self) -> TokenStream {
+        if let Some(handler) = &self.handler {
+            let protocol = &self.protocol;
+            return quote_spanned! {protocol.span()=>
+                module.associated_function(rune::runtime::Protocol::#protocol, #handler)?;
+            };
+        }
+        match self.protocol.to_string().as_str() {
+            "ADD" => quote_spanned! {self.protocol.span()=>
```
udoprog, line 132: `quote_spanned!` isn't very helpful when it's associated with large code blocks, therefore you should prefer just using `quote!` unless we've specifically added a UI test indicating there's a noted improvement in diagnostics.

crates/rune-macros/src/context.rs
```diff
@@ -111,6 +115,60 @@ pub(crate) struct FieldProtocol {
...
+                module.associated_function(rune::runtime::Protocol::ADD, |this: Self, other: Self| this + other)?;
+            },
+            "STRING_DISPLAY" => quote_spanned! {self.protocol.span()=>
+                module.associated_function(rune::runtime::Protocol::STRING_DISPLAY, |this: &Self, buf: &mut String| {
+                    use ::std::fmt::Write as _;
+                    ::std::write!(buf, "{this}")
```
udoprog, line 138: `::core::fmt::Write` and `::core::write!`.

crates/rune-macros/src/context.rs
```diff
@@ -111,6 +115,60 @@ pub(crate) struct FieldProtocol {
...
+                })?;
+            },
+            "STRING_DEBUG" => quote_spanned! {self.protocol.span()=>
+                module.associated_function(rune::runtime::Protocol::STRING_DEBUG, |this: &Self, buf: &mut String| {
+                    use ::std::fmt::Write as _;
+                    ::std::write!(buf, "{this:?}")
```
udoprog, line 144: Same here.

udoprog (commented):
Great feature, thank you!

Documentation and tests and I'd consider it done, even if we need to add more protocols later on.

The one nit I have about syntax is that I'd probably prefer the use `#[rune_derive(..)]` and `#[rune_functions(..)]` over `#[rune(protocols(..))` and `#[rune(functions(..))]`. I'm not a huge fan of nested attributes.

Regarding naming, what do you say?

crates/rune-macros/src/context.rs
ModProg: will add a ui test

ModProg:
> The one nit I have about syntax is that I'd probably prefer the use #[rune_derive(..)] and #[rune_functions(..)] over #[rune(protocols(..)) and #[rune(functions(..))]. I'm not a huge fan of nested attributes.

whatever you prefer, I don't have a strong feeling there.

> Regarding naming, what do you say?

we could also do `rune::functions` and `rune::derive`.

ModProg:
@udoprog what is your opinino on this?

> One thing to decide is whether to require the `Self::` as protocols are likely often associated functions.
> 
> This would mean that:
> 
>    1. external functions need a qualification i.e. `super::something` or similar
> 
>    2. functions in the containing module would need to be specified with `self::something`.
> 
> 
> So this would boil down, would we rather require `Self` on associated functions or `self` on non-associated functions.

udoprog:
Seems fine. For the way it's implemented you should even be able to omit `self::` if the function is unambiguous.

crates/rune-macros/src/context.rs
udoprog: FTR, don't sweat it if you can't find a difference right now, you can just prefer to use `quote!`.

ModProg:
> Seems fine. For the way it's implemented you should even be able to omit `self::` if the function is unambiguous.

currently yes the question was about changing this, i.e. making single ident function names be associated by default and only resolve paths relatively to the module.

i.e. "the_fn" would be "Self::the_fn" and "self::the_fn" would be "self::the_fn", i.e. the containing module.

Currently it is "Self::the_fn" is "Self::the_fn" and "the_fn" would resolve to "self::the_fn" i.e. the containing module.

udoprog:
Oh I see. Well, then no. Let them resolve to whatever path they're part of? They're declared within an impl that sees `Self`, so that ends up being a legal path. I don't see the reason why you'd want to make it more explicit.

crates/rune-macros/src/context.rs
```diff
@@ -113,6 +117,57 @@ pub(crate) struct FieldProtocol {
...
+                module.associated_function(rune::runtime::Protocol::STRING_DEBUG, |this: &Self, buf: &mut String| {
+                    use ::core::fmt::Write as _;
+                    ::core::write!(buf, "{this:?}")
+                })?;
+            },
+            _ => syn::Error::new_spanned(
+                &self.protocol,
+                format!("Rune protocol `{}` cannot be derived", self.protocol),
+            )
+            .to_compile_error(),
```
udoprog, line 153: Make sure to raise this error during parsing instead. I initially missed this, but I want to make sure that error handling is as coherent as possible in that it's only handled in a single place. This results in a `compile_error!` being produces in the middle of the install block which while clever is a bit surprising.

crates/rune/src/any.rs
```diff
@@ -54,6 +54,51 @@ use crate::hash::Hash;
...
+///         self.0
+///     }
+/// }
+/// 
+/// #[rune::function]
+/// fn second(it: Struct) -> usize {
+///     it.1
+/// }
+/// ```
 pub use rune_macros::Any;
```
ModProg, line 102: @udoprog The docs for the `Any` macro are actually not shown. Not sure if that is due to being reexported again at the crate root.

ModProg: AFAICT the only way around that, is moving this reexport in the crate root.

ModProg: https://github.com/rust-lang/rust/issues/81893

udoprog: Welp, time to move!

udoprog: Move the re-export and document it. Otherwise you have a circular dev-dependency for any doc tests that uses rune which rust-analyzer does not like. And all the documentation in one place.

udoprog:
PR looks to be in pretty good shape with a few things that can be done later. Would you be OK with me going over, clean a few things up and merging it?

ModProg:
> PR looks to be in pretty good shape with a few things that can be done later. Would you be OK with me going over, clean a few things up and merging it?

Sounds good.

IIRC one thing that needs fixing is the `String` type in `DISPLAY_STRING`. Either, rune needs a `rune::__private::String` that maps to the correct String dependent on if std is present. Otherwise, one could also have a feature in `rune-macros` for std.

Roba1993:
What is blocking this feature right now? I would be really interested for this functionality.

udoprog:
Rebasing, cleaning up the PR, and bandwidth.

I'll have another stint working on Rune eventually, and then I plan to pick it up again. But if you're interested, please go ahead!

PR #591, opened 2023-07-30
Add new protocol and native support for Ordering

udoprog:
I noticed while looking over https://github.com/khvzak/script-bench-rs/blob/main/benches/rune.rs that the comparison operations were implemented using associated functions, which is not very idiomatic.

This introduces the `CMP` protocol, which matches `Ord::cmp` in Rust, allowing you to implement a single handler for all total comparison operators.

At some point down the line we might want to look into supporting `PARTIAL_CMP` as well.

The second commit refines testing, and fixes support for tuple index assign operations.


ModProg:
I mean, we could also just allow `CMP` to return both `Option` or `Ordering`.

Not sure if it makes sense to separate these explicitly in an untyped language.

udoprog:
The goal is eventually to introduce gradual typing, at which point retrofitting support for partial ordering might become harder if we don't make the distinction right now. To that end I'm OK with having to look up the existence of two protocol methods to emulate `Ord: PartialOrd` if the type is unknown. Once the type is known by the compiler, it can immediately pick which one to call.

Ultimately the overriding decision making here is "whatever Rust does" as long as there's no obvious error so that we don't have to think too much about it 😄.

PR #599, opened 2023-08-08
Allow underscores at the start of a variable

### LINKED ISSUE #435: Allow variable names starting with an underscore `_`
It's sometimes useful to name variables starting with an underscore, e.g.: to temporarily silence unused warnings, prefix a numeric variable name or reflect a variable name from other source code.

I think it's fine to go with Rusts approach of only `_` being the "drop immediately" special case and other variables starting with `_` just silencing the unused warning (not sure if Rune has that already).

```rune
error: parse error
  ┌─ test.rn:2:8
  │
2 │   let _test = 0;
  │        ^^^^ expected `=`, but got ident

Error: failed to build rune sources (see diagnostics for details)
```

jbdutton:
Allow one or more underscores at the start of a variable name, like so:
```rust
fn main() {
    let _a = 0;
    _a = 42;
    _a
}
```
This changes the lexer to allow underscores at the start of an identifier, and only emits `ask::Kind::Underscore` if it's strictly `_`.

Interestingly, [this seems to match the behavior in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c1e565a5f473afbf4f1ff4472177a9ac). They both treat more than one underscore as a valid identifier. So this would work in both Rust and Rune:
```rust
fn main() {
    let ___ = 0;
    ___ = 42;
    ___
}
```
(Assuming you mark `___` as `mut` in Rust.)

Personally, I find this behavior odd. But maybe matching the Rust compiler behavior is best?

Fixes #435.

### DIFF
--- a/crates/rune/src/ast/ident.rs
+++ b/crates/rune/src/ast/ident.rs
@@ -6,6 +6,7 @@ fn ast_parse() {
 
     rt::<ast::Ident>("foo");
     rt::<ast::Ident>("a42");
+    rt::<ast::Ident>("_ignored");
 }
 
 /// An identifier, like `foo` or `Hello`.
--- a/crates/rune/src/parse/lexer.rs
+++ b/crates/rune/src/parse/lexer.rs
@@ -143,10 +143,17 @@ impl<'a> Lexer<'a> {
             self.iter.next();
         }
 
-        let (ident, span) = self.iter.source_from(start);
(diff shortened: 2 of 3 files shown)


udoprog:
Short and sweet. Thank you!

The msrv error is not your fault, it's an upstream dependency I'll have to deal with once I bake a new release.

PR #643, opened 2023-10-07
Add support for coercing bevy mutable references to values

MalekiRe:


udoprog:
I fiddled a bit with the patch, so please take it for a spin.

Some notes:
I like keeping everything in `bevy_related`. I also noted that `AnyObjVTable` is supposed to be public (but wasn't). So I've made it public.

AFAICT, using a trait `UnsafeToValue2` doesn't net you much. You just have to pull it into scope anyways.

The reason we can't implement `UnsafeToValue` (which would be nice) is because of the `ToValue` blanket implementation. I'd definitely consider the blanket implementation because it would enable this which makes calling into rune better for you. This shouldn't be a big deal, since `Any` is mostly derived so we can easily just implement this as appropriate.

`MutUntyped` has lifetimes, so `MutUntyped` is the same as `MutUntyped<'static>`. Coercing to `'static` is a bit sketchy, but without bevy exposing raw APIs (which I think they should to enable sound uses like this) we don't have much of a choice.

PR #649, opened 2023-10-23
Make Formatter functions public

Roba1993:
The Formatter is needed for various interactions with a rune values. Therefore I need to be able to get a string as result out of the Formatter.

Why is this needed?
When receiving a generic rune::Value from the vm, this value can be a Result. To debug format the error result in rust, the formatter is now needed. Unfortunately the value can't be received out of the Formatter as of now.

### DIFF
--- a/crates/rune/src/runtime/fmt.rs
+++ b/crates/rune/src/runtime/fmt.rs
@@ -37,7 +37,7 @@ impl Formatter {
     }
 
     #[inline]
-    pub(crate) fn with_capacity(capacity: usize) -> alloc::Result<Self> {
+    pub fn with_capacity(capacity: usize) -> alloc::Result<Self> {
         Ok(Self {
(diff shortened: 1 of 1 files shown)


udoprog:
`Formatter` internals should not need to be public, instead you're expected to do something like this if you need to debug print something:

```rust
let mut vm: Vm = todo!();
let value: rune::Value = todo!();
vm.with(|| format!("{:?}", value));
```

The `Formatter::new` constructor probably isn't supposed to be public either :sweat_smile:

Roba1993:
Seems quite an overhead for me to be honest. The rune::Value result need to be put again into an VM to format it to text 😕

udoprog:
You have to run it through a vm regardless, formatting is customized through a protocol and will need access to native functions if a native thing is being formatted. If it doesn't have access to a vm and needs it, it would panic.

PR #651, opened 2023-10-25
Allow async cli execution

Roba1993:
Allows async cli execution. The sync version spawns a Tokio instance which panics, when Tokio is already spawned before.


udoprog:
LGTM, cheers!

PR #652, opened 2023-10-30
Make function for formatting of Source Code public

Roba1993:


udoprog:
Not that I oppose it, but any particular reason you want this public.

It will probably change a fair bit by next release.

Roba1993:
Within the exolynk platform the code should be automatically formatted while saving & testing. The code is inside the db, so formatting a file is not possible.

Roba1993:
Implemented by [#79e81a0](https://github.com/rune-rs/rune/commit/79e81a02d01ac8d40fa8c7b595bd7262282dd552)

PR #654, opened 2023-11-02
vscode: pass extra env when validating

SWW13:
The extra env variables were only passed for launching the language server not when validating the executable is working.

I also fixed JS warnings for non type-safe compares.


udoprog:
Nice, thank you!

PR #665, opened 2024-02-20
Create a public function for source formatting

cocuh:
Alternative solution for #652.

In most cases, users don't need FormattingError enum itself, and just serializing the error message makes things easier in later code chagnes I believe.

### DIFF
--- a/crates/rune/src/fmt.rs
+++ b/crates/rune/src/fmt.rs
@@ -9,6 +9,7 @@ mod indent_writer;
 mod printer;
 mod whitespace;
 
+use crate::alloc::String;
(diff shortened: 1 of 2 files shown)


udoprog:
You don't have to use anyhow to hide the error enum. A public struct with a private field implementing std::error::Error would achieve the same result without the feature test.

udoprog:
Woops, the auto merged triggered prior to building 🤔. No big deal, I'll fix it when I get around to it.

Thanks!

cocuh:
Ooop, it seems my commit makes build failure. Let me create PR for the fix.

PR #669, opened 2024-02-28
Add String::{to_lowercase,to_uppercase}

dstoza:


crates/rune/src/modules/string.rs
```diff
@@ -1103,5 +1105,87 @@ fn parse_char(s: &str) -> Result<char, char::ParseCharError> {
...
+/// Languages without case are not changed:
+///
+/// ```rune
+/// let new_year = "农历新年";
+///
+/// assert_eq!(new_year, new_year.to_lowercase());
+/// ```
+#[rune::function(instance)]
+fn to_lowercase(s: &str) -> VmResult<String> {
+    VmResult::Ok(vm_try!(String::try_from(s.to_lowercase())))
```
udoprog, line 1149: This uses `to_lowercase` provided by `std`, which performs internal allocations using the default system allocator. In Rune we have to use types provided by [`rune-alloc`](https://docs.rs/rune-alloc) to ensure that memory limits are enforced.

In practice, I think this maybe means something like this:

```rust
let mut out = String::try_with_capacity(s.len());

for c in s.chars() {
    for c in c.to_uppercase() {
        output.push(c);
    }
}

VmResult::Ok(output)
```

dstoza:
A least `to_uppercase` was easy 🙃. I decided to inline some code from `core` rather than either (a) taking an unstable dependency (which requires a crate-level opt-in) or (b) not supporting Greek correctly, but I'm open to other possible solutions.

udoprog:
Thank you for the PR!

PR #672, opened 2024-02-29
Add Value::borrow_any_{ref,mut} to get a typed reference without ownership

dstoza:


crates/rune/src/runtime/value.rs
```diff
@@ -1335,6 +1335,23 @@ impl Value {
         VmResult::Ok(vm_try!(Future::from_value(value)))
     }
 
+    /// Get the value as a typed value.
+    #[inline]
+    pub fn as_any<T>(&self) -> Result<BorrowRef<'_, T>, RuntimeError>
```
udoprog, line 1340: Naming when there's this many methods is tricky, and something I'll definitely want to go over before the next release. The pattern with the macros currently is:

* `borrow_<type>_ref`
* `borrow_<type>_mut`
* `into_<type>_ref`
* `into_<type>_mut`

But this doesn't mesh well with `any`. But maybe this as a result should be `borrow_any_ref`?

udoprog:
Thanks!

PR #680, opened 2024-03-21
Minor LSP fixups

tgolsson:
* A formatting error in a detail
* I started writing a rune mode for emacs (again...), and noticed that returning a u64 as the data is out of spec -- only u32/i32 are legal. This causes emacs to barf because the JSON parser there assumes `i64` as the integer value. This data doesn't seem to be used, but at least this transform is reversible if we want it.
* I noticed that completing `json::` didn't include `json::` as the prefix, leading to very odd results. I think the symbol-heuristic is a bit too naive still, but this is a tiny bit better. For example, completing the example would lead to `json::json::from_string`, whereas now it'll correctly only have `json::to_string`


udoprog:
Thanks for giving this some attention!

PR #682, opened 2024-03-28
Fallback to debug-printing generic info

### LINKED ISSUE #681: `tracing::instrument` on `op_return_internal` is broken
Noticed this when upping the log-level on an old project. If the return value doesn't have `Protocol::DEBUG_STRING` it'll just hard-crash when the return happens. Took quite a while to figure out why... :D

https://github.com/rune-rs/rune/blob/fc2bc0c6b48893b819dc1681ab08a19655ce77b1/crates/rune/src/runtime/vm.rs#L2029-L2035

I can point-fix it, but I think in general it'd be nicer if dbg! etc was less prone to exploding. Maybe an optional feature to "fall back" to something like `<%T object at %p>`, which isn't what most people would want but won't lead to spurious BOOM.

tgolsson:
Fixes #681 by falling back to a `<$OBJECT_NAME at $ADDRESS>` style of debug.

### DIFF
--- a/crates/rune/src/runtime/value.rs
+++ b/crates/rune/src/runtime/value.rs
@@ -851,8 +851,9 @@ impl Value {
                 vm_write!(f, "{:?}", value);
             }
             _ => {
+                // reborrow f to avoid moving it
                 let result =
-                    caller.call_protocol_fn(Protocol::STRING_DEBUG, self.clone(), (&mut *f,)); // reborrow f to avoid moving it
+                    caller.call_protocol_fn(Protocol::STRING_DEBUG, self.clone(), (&mut *f,));
 
                 if let VmResult::Ok(result) = result {
                     vm_try!(<()>::from_value(result));
--- a/crates/rune/src/runtime/value.rs
+++ b/crates/rune/src/runtime/value.rs
@@ -858,7 +858,7 @@ impl Value {
                     vm_try!(<()>::from_value(result));
                 } else {
                     let type_info = vm_try!(self.type_info());
-                    vm_write!(f, "<{} object at {:p}>", type_info, self);
+                    vm_write!(f, "<{} object at {:p}>", type_info, self.inner);
                 }
             }
         };
--- a/crates/rune/src/runtime/value.rs
+++ b/crates/rune/src/runtime/value.rs
@@ -852,10 +852,14 @@ impl Value {
             }
             _ => {
                 let result =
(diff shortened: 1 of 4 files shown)


crates/rune/src/runtime/value.rs
```diff
@@ -852,10 +852,14 @@ impl Value {
...
-                    vm_try!(caller.call_protocol_fn(Protocol::STRING_DEBUG, self.clone(), (f,)));
+                    caller.call_protocol_fn(Protocol::STRING_DEBUG, self.clone(), (&mut *f,)); // reborrow f to avoid moving it
 
-                vm_try!(<()>::from_value(result));
-                return VmResult::Ok(());
+                if let VmResult::Ok(result) = result {
+                    vm_try!(<()>::from_value(result));
+                } else {
+                    let type_info = vm_try!(self.type_info());
+                    vm_write!(f, "<{} object at {:p}>", type_info, self);
```
udoprog, line 861: Not a bad approach in general, but I am weary of potentially leaking real address information. In itself it shouldn't be usable for anything nefarious, but might maybe be used to defeat e.g. [address space randomization](https://en.wikipedia.org/wiki/Address_space_layout_randomization). Maybe we should stick to type info?

tgolsson: I wanted something to make instances a bit more unique, but also the `self` really isn't interesting here since the inner value is a arc-like thing.

Do we have any other info to separate something like `dbg!(foo_a, foo_b);`?

udoprog: Ah, so `self` points to the `Value` struct, and by extension wherever that is stored (the program stack, the vm stack, ...). What you probably meant to do would be to print the address of the pointer stored in [`Shared<ValueKind>`](https://github.com/rune-rs/rune/blob/fc2bc0c6b48893b819dc1681ab08a19655ce77b1/crates/rune/src/runtime/shared.rs#L21)? At least that address should be harder to use to locate interesting things in memory.

If that's also sensitive we could also opt to allocate pseudo-random identifiers (through e.g. a static AtomicUsize) and store them in `SharedBox`. Either enabled through a feature or a debug build.

tgolsson: Switched it to use the pointer from self.inner, as a start. I can't claim to care about the security implications  of it still, so I'll defer to you if you'd prefer some other approach. The downside of using a feature/debug/... is that it doesn't have a fallback. I guess we could opt-out of printing addresses with a feature though, or not-in-release.

PR #685, opened 2024-04-05
Generate warnings when using deprecated functions

Roba1993:
This PR extends rune with the functionality to emit warnings when deprecated functions are used.
When possible the deprecation warning is created while compilation.
For instance and field functions the warning will only be created/emitted when the VM has a shared Diagnostic struct set. If this diagnostic has not been set, the runtime performance will not be affected.


udoprog (commented):
Hey, and thank you for your contribution!

#### More generic appraoch

I'd definitely like to see `Vm` have a more generic mechanism for emitting *stuff* that we can generate diagnostics of, preferably one which by default compiles to nothing.

Strictly speaking, what it seems like you need is to know which type hashes have been accessed, which should require that we embed a complete diagnostics object inside of the `Vm`. Instead, calling a trait like this should be enough:

```
trait VmDiagnostic {
    fn function_used(&mut self, hash: Hash);
}
```

You can then provide an implementation of that trait to collect that diagnostics. Or one can be provided by default in the form of a `&mut Diagnostics`.

#### Nested VMs

I don't think the current approach works with nested vms which are used in at least these instances:
* A future being awaited.
* A closure being constructed and called.

These include virtual machines which have their own stack and state, so I think the diagnostics which is currently associated with the parent machine would be lost when they are used.

The most surefire way to deal with this would be to use thread-locals for the instrumentation. There already is one in the ["env" module](https://github.com/rune-rs/rune/blob/main/crates/rune/src/runtime/env.rs#L45) you might want to extend. It could for example include a pointer to an optional trait object like `Option<NonNull<dyn VmDiagnostic>>`.

What you'd need to somehow provide a reference wherever `env::Guard` is constructed [like here](https://github.com/rune-rs/rune/blob/main/crates/rune/src/runtime/vm.rs#L2962). I honestly wouldn't mind if this was done as an additional `fn run` variant, like `fn run_with_diagnostics(&mut self, diagnostics: &mut dyn VmDiagnostics)`.

Roba1993:
Hey, thank you for the feedback. 

I'm trying to implement it right now. But I have the following questions along that way:
Adding `Option<NonNull<dyn VmDiagnostic>>` to the Env sounds okay, but I need to have the VmDiagnostic mutable, so that I can add the warnings to it. Would an `Option<Arc<Mutex<dyn VmDiagnostic>>>` not be better for that case? Or is there another way I'm not seeing, like implementing `VmDiagnostic` for Arc<Mutex<Diagnostics>>?

Also where would be the best places to call `function_used(&mut self, hash: Hash)` then? The same places as of now and receive the correct Env with `self::no_std::rune_env_get();`?

udoprog:
> I'm trying to implement it right now. But I have the following questions along that way: Adding `Option<NonNull<dyn VmDiagnostic>>` to the Env sounds okay, but I need to have the VmDiagnostic mutable, so that I can add the warnings to it. Would an `Option<Arc<Mutex<dyn VmDiagnostic>>>` not be better for that case? Or is there another way I'm not seeing, like implementing `VmDiagnostic` for Arc<Mutex>?

`VmDiagnostic` can be implemented for any type, including like `SomeWrapper`. Exactly how that is done would be a local decision for anyone who wants to collect diagnostics.

See [this skeleton as an example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=26e5fd60941e953ee8be78e7503220b0).

> Also where would be the best places to call `function_used(&mut self, hash: Hash)` then? The same places as of now and receive the correct Env with `self::no_std::rune_env_get();`?

Yeah, that's seems about right.

Roba1993:
Thank you for the explanation. My problem was that I didn't know, that `Option<NonNull<dyn VmDiagnostic>>` is like a `&mut dyn VmDiagnostic`. I haven't worked with unsafe so far....

The last problem right now is, how to execute the code now with the diagnostics attached. In my code where I use rune, I call all my functions like this:
```rust
// set function to call with it's parameters to the vm
let execution = match vm.send_execute([name], args.into()) {
   Ok(vm) => vm,
   Err(e) => {
      return CodeExectionResult::from_error(
         start_time,
         io,
         e,
         Some(diagnostics_writer),
         diagnostics,
         instructions,
      );
   }
};

// start the execution in an async way
let future = execution.async_complete();
```

My problem is, that I don't see the guard set up for the `async_complete()` function here: https://github.com/rune-rs/rune/blob/37e4bd0a86da6012323ce26b19bbb3f1869dde05/crates/rune/src/runtime/vm_execution.rs#L446

udoprog:
Easiest would probably be to pass it in as a parameter through set of functions (`**_with_diagnostics`) which allow you to pass in a `&mut dyn VmDiagnostics`.

Alternatively they can be set externally through the same thread-locals, where the first call would just modify the diagnostics hook.

I think I'd prefer option 1 though for now. Unless there is a cleaner option. It is dead simple, super explicit, and a great position to document what the parameter actually does.

Roba1993:
To be really honest, I don't really like the first approach. If I want to support an sync and async version I would completely copy the run function as you said here:
https://github.com/rune-rs/rune/blob/37e4bd0a86da6012323ce26b19bbb3f1869dde05/crates/rune/src/runtime/vm.rs#L2959

Afterwards also nearly all functions from the VmExecution:
https://github.com/rune-rs/rune/blob/main/crates/rune/src/runtime/vm_execution.rs#L278

And then make the specific functions for the call functions:
https://github.com/rune-rs/rune/blob/37e4bd0a86da6012323ce26b19bbb3f1869dde05/crates/rune/src/runtime/vm.rs#L388
https://github.com/rune-rs/rune/blob/37e4bd0a86da6012323ce26b19bbb3f1869dde05/crates/rune/src/runtime/vm.rs#L429

Either I copy all this code or pollute them with a diagnostic parameter. Both is not really nice from my perspective...

udoprog:
Broadly speaking what you could to do for option 1 is move the real implementation into a method which takes an optional diagnostics reference `Option<&mut dyn VmDiagnostics>`, then keep the existing ones calling that private function with `None`, and the new functions call it with `Some(diagnostics)`.

Then you'd just have to refactor the existing functions to thread the parameter through and build two facades.

crates/rune/src/runtime/vm_execution.rs
```diff
@@ -206,14 +208,17 @@ where
...
-    async fn inner_async_resume(&mut self) -> VmResult<GeneratorState> {
+    async fn inner_async_resume(
+        &mut self,
+        diagnostics: Option<&mut dyn VmDiagnostics>,
+    ) -> VmResult<GeneratorState> {
         loop {
             let vm = self.head.as_mut();
 
-            match vm_try!(vm.run().with_vm(vm)) {
+            match vm_try!(vm.run(diagnostics).with_vm(vm)) {
```
Roba1993, line 221: This fails, because I moved it before...

crates/rune/src/runtime/vm_execution.rs
```diff
@@ -272,15 +277,18 @@ where
...
+    fn inner_resume(
+        &mut self,
+        diagnostics: Option<&mut dyn VmDiagnostics>,
+    ) -> VmResult<GeneratorState> {
         loop {
             let len = self.states.len();
             let vm = self.head.as_mut();
 
-            match vm_try!(vm.run().with_vm(vm)) {
+            match vm_try!(vm.run(diagnostics).with_vm(vm)) {
```
Roba1993, line 291: This should also fail, but is not right now in my compiler....

Roba1993 (commented):
This is my mutable Option problem, which I still have. Everything else works fine so far...

crates/rune/src/runtime/runtime_context.rs
```diff
@@ -34,16 +35,20 @@ pub struct RuntimeContext {
     functions: hash::Map<Arc<FunctionHandler>>,
     /// Named constant values
     constants: hash::Map<ConstValue>,
+    /// Deprecation messages for functions
+    deprecations: hash::Map<Box<str>>,
```
udoprog, line 39: The deprecation messages should not be associated with the runtime context. We can avoid storing them here and instead have the emitted diagnostics be decorated if a `Unit` (stored in [`DebugInfo`](https://docs.rs/rune/latest/rune/runtime/unit/struct.Unit.html#method.debug_info)) and `Context` is provided.

Else, if they're not available, you'd print a more generic message like:

```
Called deprecated function with hash 0xdeadbeef
```

Note that we have to do something like that regardless (merge deprecation messages from different sources), to support deprecation messages from units which cannot be stored in the runtime context.

crates/rune/src/runtime/vm_diagnostics.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl VmDiagnostics for Diagnostics {
+    fn function_used(
+        &mut self,
+        context: &Arc<RuntimeContext>,
+        unit: &Arc<Unit>,
+        hash: Hash,
+        at: usize,
+    ) -> VmResult<()> {
```
udoprog, line 30: This interface seems more complicated than it has to be.

In my mind all we need to emit is a `hash` and an instruction pointer `at`. Correlating the diagnostic can be done done external to the trait implementation.

crates/rune/src/diagnostics/warning.rs
```diff
@@ -114,6 +117,14 @@ pub enum WarningDiagnosticKind {
         /// Span where the semi-colon is.
         span: Span,
     },
+    UsedDeprecated {
+        /// The span which is deprecated
+        span: Span,
+        /// The context in which it is used.
+        context: Option<Span>,
+        /// The defined deprecation message
+        message: Box<str>,
```
udoprog, line 126: The message doesn't have to be stored in here, it can instead be looked up in the `Context` or `Unit` when diagnostics is being emitted.

crates/rune/src/compile/context.rs
```diff
@@ -881,12 +889,16 @@ impl Context {
         &mut self,
         hash: Hash,
         handler: &Arc<FunctionHandler>,
+        deprecation: &Option<Box<str>>,
```
udoprog, line 892: Nit: Should probably prefer using `Option<&str>`. There's no benefit for `try_clone` to know it's a `Box<str>`. And we have [`TryToOwned`](https://docs.rs/rune-alloc/latest/rune_alloc/borrow/trait.TryToOwned.html#impl-TryToOwned-for-str).

crates/rune/src/compile/context.rs
```diff
@@ -305,6 +308,11 @@ impl Context {
         indexes.iter().map(|&i| &self.meta[i])
     }
 
+    /// Lookup deprecation by function hash.
+    pub(crate) fn lookup_deprecation(&self, hash: Hash) -> Option<&Box<str>> {
```
udoprog, line 312: Nit: Same: `Option<&str>`.

udoprog (commented):
A few things to change, but really coming together now. Thanks for putting up with me!

crates/rune/src/runtime/vm_execution.rs
```diff
@@ -181,6 +183,22 @@ where
...
     }
 
+    /// Complete the current execution without support for async instructions.
+    ///
+    /// If any async instructions are encountered, this will error. This will
+    /// also error if the execution is suspended through yielding.
+    pub fn complete_with_diagnostics(
+        &mut self,
+        diagnostics: Option<&mut dyn VmDiagnostics>,
+    ) -> VmResult<Value> {
```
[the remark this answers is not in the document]
Roba1993: I have done it, but then I saw a lot of test and other code is using this function directly. I switched it back, to be more backwards compatible. But I can also remove this specific `_with_diagnostics` function if you prefer the breakage.

crates/rune/src/runtime/vm_diagnostics.rs
Roba1993: I removed the `context` and I totally agree with you on this. The `unit` would be handy, to be able to create a warning, which has the span directly attached to it and stay compatible with the other warnings.
If you still prefer to remove the unit here, I would make the warning with a `Option<Span>` and in addition `Option<usize>` for the ip location.
The only problem I see then, is how I can handle this https://github.com/rune-rs/rune/blob/8e84d521e9146e45639127f048cd029a7b7b5a59/crates/rune/src/diagnostics/warning.rs#L49 when no span is defined....

crates/rune/src/runtime/vm_execution.rs
```diff
@@ -181,6 +183,22 @@ where
...
+            GeneratorState::Complete(value) => VmResult::Ok(value),
+            GeneratorState::Yielded(..) => VmResult::err(VmErrorKind::Halted {
+                halt: VmHaltInfo::Yielded,
+            }),
+        }
+    }
+
     /// Resume the current execution with the given value and resume
     /// asynchronous execution.
     pub async fn async_resume_with(&mut self, value: Value) -> VmResult<GeneratorState> {
```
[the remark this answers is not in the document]
Roba1993: A builder pattern would be a lot better I'm with you. To be honest right now it's also a bit confusing for people to understand the difference between an `Vm`, `VmExecution` & `VmSendExecution` when starting to integrate run into won projects.

crates/rune/src/diagnostics/emit.rs
```diff
@@ -475,6 +535,63 @@ where
...
+where
+    O: WriteColor,
+{
+    let mut notes = ::rust_alloc::vec::Vec::new();
+    let mut labels = ::rust_alloc::vec::Vec::new();
+    let mut message = String::new();
+
+    match this.kind {
+        RuntimeWarningDiagnosticKind::UsedDeprecated { hash } => {
+            // try to get the function name - this needs to be improved
```
Roba1993, line 556: This works most of the time, but not for instance fields. `DebugInfo` only holds the information from source files and not static generated infos. This is okay for me at this time but maybe you have a nice idea.

crates/rune/src/languageserver/state.rs
```diff
@@ -590,6 +590,9 @@ fn emit_scripts(
             Diagnostic::Warning(e) => {
                 report(build, reporter, e.source_id(), e, to_warning)?;
             }
+            Diagnostic::RuntimeWarning(w) => {
+                todo!()
```
Roba1993, line 594: My question here is, di you want to show it at all? The actual `report` function is not compatible at all right now...

Roba1993 (commented):
From my perspective, we are quite close to the goal. Can you answer this two questions and the the last two ones from your code review?

crates/rune/src/languageserver/state.rs
udoprog: A language server will never execute the code, so will never see runtime diagnostics. Another reason why it's better to collect runtime diagnostics into a separate type, like `RuntimeDiagnostics`.

But at the very least don't leave the panic. This should either emit a warning, error in a recoverable way, or be ignored. We don't want the LS to ever crash.

Roba1993:
Please let me know, what is still blocking for a merge.

crates/rune/src/tests/deprecation.rs
```diff
@@ -61,17 +61,15 @@ fn check_diagnostic(context: &Context, diagnostic: &Diagnostic, msg: &str) {
...
         Diagnostic::RuntimeWarning(w) => match w.kind() {
             RuntimeWarningDiagnosticKind::UsedDeprecated { hash, .. } => {
                 let message = context.lookup_deprecation(*hash);
                 assert_eq!(message, Some(msg));
             }
-            #[allow(unreachable_patterns)]
-            _ => panic!("Expected a UsedDeprecated warning"),
         },
-        _ => panic!("Diagnostic need to be of typ warning"),
+        kind => panic!("Unexpected diagnostics: {kind:?}"),
```
udoprog, line 72: In case this assertion fails, we'll probably need to know how it fails to fix the error.

There is a better method available called assert_errors! which does a lot of this by default, but it doesn't cleanly plug into how you're testing the context.

crates/rune/src/diagnostics/emit.rs
```diff
@@ -122,8 +122,8 @@ impl Diagnostics {
         &self,
         out: &mut O,
         sources: &Sources,
-        debug_info: Option<&DebugInfo>,
-        context: Option<&Context>,
+        unit: &Unit,
+        context: &Context,
```
udoprog, line 126: I've opted to just pass in the `Unit` and `Context` directly. If these are not available, you'd use `emit`. But ultimately this should probably be made into a builder which allows you to more freely associate build elements that are available.

udoprog (approved):

crates/rune/src/diagnostics/emit.rs
udoprog: Eh, that's tough. We'd have to emit a different deprecated method which includes the type hash so that we can reverse the hash, like [we do here](https://github.com/rune-rs/rune/blob/25b256ee06a6d3cbd5e45f647c7d809f7fb3f8f6/crates/rune/src/diagnostics/emit.rs#L206).

It would require a bit more work, and would be an OK future change.

udoprog:
So I looked over this some more, and the way the APIs worked had potential to cause a soundness issue:

If `env::with` was called recursively (which it might be) it would cause multiple exclusive references to `&mut dyn VmDiagnostics` to be live on the stack. This wouldn't be OK. So I've split it into two methods - on which only accesses shared parts, and the other which accesses exclusive parts. The exclusive one cannot be called recursively.

PR #691, opened 2024-04-18
Added bytes support for http

Roba1993:
This adds byte support for the rune http response struct and allows no-cors request for WASM targets.


udoprog:
Looks good!

* You'll have to register the functions in the `fn module` to export them.
* It would be great if you added `no_run` doc tests. I'll be enforcing a policy that everything has examples once everything has examples.

udoprog:
Thank you!

PR #706, opened 2024-05-07
make http Client public so it can be instantiated in Rust

tksfz:


udoprog:
I will need a better rationale for this. Currently the client is pretty much unusable outside of Rune since no focus has been put into making it compatible with Rust. So closing for now.

PR #708, opened 2024-05-19
Make equality symmetric for user defined types.

Deukhoofd:
As a user of Rune, I'd like equality to be symmetric, so that the order of the equality does not matter. This means that ``a == b`` implicitly also means that ``b == a``. 

Currently this is not the case when defining an associated method for a type, where the other type is not the same as the instance type. An example of the use case for this is a user defined string wrapper, where we want to implement equality with a Rune string, and want it to work both ways. In the current implementation, if we implement equality from type a to type b, we can do a == b, but when doing b == a we get an error that the function does not exist, as it's only implemented on type a.

This pull request aims to fix that.

To do so the following changes were made:
- The protocol callers were changed to have their implementation inside of ``try_call_protocol_fn``, instead of ``call_protocol_fn``. ``try_call_protocol_fn`` already existed, but due to its implementation could never return a CallResult that was not ``CallResult::Ok``. After this change, it's now possible to receive a ``CallResult::Unsupported`` from ``try_call_protocol_fn``. 
- The  ``call_protocol_fn`` now calls ``try_call_protocol_fn``, and returns a ``VmErrorKind::MissingProtocolFunction`` error if it receives a ``CallResult::Unsupported``. Note that this is different from the previous error, which was ``VmErrorKind::MissingFunction``. The ``MissingProtocolFunction`` error felt like a far clearer error here.
- The ``partial_eq_with`` and ``eq_with`` functions of ``Value`` now try to resolve ``b == a``  if they're unable to resolve ``a == b``. 

As a side effect, due to the above changes where ``try_call_protocol_fn`` can now return values other than ``CallResult::Ok`` or VmError, it's now possible to reach several errors that were unreachable before. For cmp, eq, partial_eq, and hash, you'll now receive the already existing UnsupportedBinaryOperation or UnsupportedUnaryOperation errors when the protocol is not implemented.

### DIFF
--- a/crates/rune/src/runtime/protocol_caller.rs
+++ b/crates/rune/src/runtime/protocol_caller.rs
@@ -13,8 +13,16 @@ pub(crate) trait ProtocolCaller: Sized {
         target: Value,
         args: A,
     ) -> VmResult<Value>
-    where
-        A: GuardedArgs;
+        where
+            A: GuardedArgs {
+        match vm_try!(self.try_call_protocol_fn(protocol, target, args)) {
+            CallResult::Ok(value) => VmResult::Ok(value),
+            CallResult::Unsupported(instance) => VmResult::err(VmError::from(VmErrorKind::MissingProtocolFunction {
+                protocol,
+                instance: vm_try!(instance.type_info()),
+            })),
+        }
+    }
 
     /// Call the given protocol function.
     fn try_call_protocol_fn<A>(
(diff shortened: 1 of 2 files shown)


udoprog:
I appreciate you taking the time suggesting this!

This is probably not something I'd want to see solved at runtime, because it introduces an asymmetry in behavior which is not present at compile time in Rust.

Broadly speaking, if a design space is open, rather than investigate it I defer to "whatever Rust does", and in terms of looking up equality implementations, Rust [will treat it as a two-argument operator](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=568e25ff9b8c98991597c9e2ac240086). Unfortunately this would be too much of a deviation from that standard for me.

However, nothing prevents you from introducing high-level support which ensures symmetry when registering operators, such as when registering them in a module to ensure that both left / right and right / left operators are registered.

Issue #684, opened 2024-04-03
`rune fmt` breaks multiline comments


SWW13:
`error.rn`:
```
pub fn main() {
    /*
    test
    */
}
```

```
$ cargo install rune-cli --version 0.13.2
$ rune fmt error.rn
$ cat error.rn
pub fn main() {
/*
    test    *
}
```

I have experienced the issue of disappearing lines for a while now and just now triaged the source, ~~so it may be related to the changes in #527 / #539.~~
I further tracked it down, it was already present in `0.13.0`, the initial commit `802101c9` returned a parser error for the file:
```
$ cargo +1.74.0 run --bin rune -- fmt test.rn
!! error.rn: error while parsing source
0 unchanged
0 succeeded
1 failed
```

Issue #703, opened 2024-04-29
`rune fmt` removes indent of comments


SWW13:
`cargo install rune-cli --version 0.13.2`

`test.rn`:
```rust
pub fn main() {
    if true {
        // test
    }
}
```

`rune fmt test.rn`

```rust
pub fn main() {
    if true {
    // test
    }
}
```

Issue #693, opened 2024-04-23
`VecDeque` can't be sent through Rune


HoloTheDrunk:
**Attempt**
```rust
#[derive(Any)]
struct MyStruct {
    #[rune(get)]
    dvec: VecDeque<i64>,
}
```

**Expected behaviour**
The code compiles without issue.

**Actual behaviour**
The compilers throws this error:
```
error[E0277]: the trait bound `{closure@src/main.rs:8:10: 8:13}: InstanceFunction<_, Plain>` is not satisfied
    --> src/main.rs:8:10
     |
8    |   #[derive(Any)]
     |            ^^^ the trait `InstanceFunction<_, Plain>` is not implemented for closure `{closure@src/main.rs:8:10: 8:13}`
9    |   struct MyStruct {
10   | /     #[rune(get)]
11   | |     dvec: VecDeque<i64>,
     | |_______________________- required by a bound introduced by this call
     |
note: required by a bound in `rune::Module::field_function`
    --> /home/RDuhen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rune-0.13.2/src/module/module.rs:1603:12
     |
1595 |     pub fn field_function<N, F, A>(
     |            -------------- required by a bound in this associated function
...
1603 |         F: InstanceFunction<A, Plain>,
     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Module::field_function`
```

**Note**
Another error message I have not been able to reproduce said something along the lines of "[`VecDeque`](https://rune-rs.github.io/docs/std/collections/VecDeque.struct.html) is missing a Rune `clone` implementation", which could be what's prevent passing `VecDeque`s to a Rune VM.

udoprog:
I think it's just lacking `FromValue` and `ToValue` implementations.

You'd also be encouraged to use [`rune::alloc::vec_deque`](https://docs.rs/rune/latest/rune/alloc/vec_deque/index.html), and the mapping would go to the [`VecDeque` in the `collections` module](https://github.com/rune-rs/rune/blob/main/crates/rune/src/modules/collections/vec_deque.rs).

HoloTheDrunk:
Ah, using `rune::alloc::VecDeque` leads to the `Clone` error I couldn't get again.
```rust
error[E0277]: the trait bound `rune::alloc::VecDeque<i64>: Clone` is not satisfied
 --> src/main.rs:8:10
  |
8 | #[derive(Any)]
[...]

PR #710, opened 2024-06-13
Format Fixes

SWW13:
This pull requests fixes issue #684, #693 and #703 and adds multiple tests to ensure formatting works as expected.

I also tested against all rune files in the project to make sure no new regression is introduced.

### DIFF
--- a/crates/rune/src/fmt/comments.rs
+++ b/crates/rune/src/fmt/comments.rs
@@ -114,7 +114,7 @@ fn parse_block_comment(chars: &mut CharIndices<'_>) -> Option<usize> {
     while let Some((_, c)) = chars.next() {
         if c == '*' {
             if let Some((_, '/')) = chars.clone().next() {
-                return Some(chars.next()?.0);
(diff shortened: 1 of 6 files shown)


udoprog:
Hey, Thanks for fixing these!

Rebased since a new stable release happened with a few new changes.

Note that `nightly` is [expected to be broken](https://github.com/rust-lang/rust/issues/123748#issuecomment-2168055393) for now. It is just a canary.

udoprog:
Got a legit test failure I've not dug into: https://github.com/rune-rs/rune/actions/runs/9517091651/job/26235234512?pr=710#step:6:1174

SWW13:
That's due to this change: https://github.com/rune-rs/rune/pull/710/commits/2d4da8e335e343fe055ad44ce98ce8e3596482eb

Previously the trailing `/` was missing, so I assumed the span end must be after the char for it to be included.

udoprog:
Very nice, thank you!

Issue #640, opened 2023-10-04
Feature request: Support autocomplete for Ace editor


Roba1993:
With the new documentation feature in development (doc in html) this information could also be extracted to build autocompletion for the ace editor.
This could be used for:
- The rune website to better show what a user can do
- Custom web implementation who want to provide users an autocomplete

PR #714, opened 2024-06-18
Support autocomplete for Ace editor

Roba1993:
This PR adds the support for generating autocomplete instructions for the ace editor.
Refers to this feature request: https://github.com/rune-rs/rune/issues/640

One topic is to detect `std` imports, which are available globally, like `std::vec::Vec` towards `Vec`.

Generate the autocomplete with the following rune cli command:
```bash
rune ace --output your/location/folder --extensions
```
The optional output parameter defines a different folder where the autocomplete.js file will be created.
The extensions flag will add a `?` to all Option/Result returning functions and a `.await` to all async functions.

How to use this within your ace setup:
```js
// ensure to register this ace file as dependency
//<script src="../build/src-noconflict/ext-language_tools.js"></script>

// import the generated autocomplete.js
import runeCompleter from "rune-ace/autocomplete";

// setup your ace editor as normal
var editor = ace.edit("editor");
editor.setTheme("ace/theme/tomorrow");
editor.session.setMode("ace/mode/html");

// enable autocompletion and optional snippets
editor.setOptions({
  enableBasicAutocompletion: true,
  //enableSnippets: true,
  enableLiveAutocompletion: true
});

// register the rune autocomplete
let langTools = ace.require("ace/ext/language_tools");
langTools.addCompleter(runeCompleter);
```

### DIFF
--- a/crates/rune/src/doc/autocomplete.rs
+++ b/crates/rune/src/doc/autocomplete.rs
@@ -178,6 +178,20 @@ impl<'a> AutoCompleteCtx<'a> {
 
     fn get_fn_param(f: &Function) -> Result<String> {
         let mut param = String::try_from("(")?;
+
+        // add arguments when no argument names are provided
+        if let Some(args) = f.args {
+            if args != 0 && f.args != f.arg_names.map(|a| a.len()) {
+                for i in 0..args {
+                    if i != 0 {
+                        param.try_push_str(", ")?;
+                    }
+                    param.try_push_str("value")?;
+                }
+            }
+        }
+
+        // add argument names
         for a in f.arg_names.into_iter().flatten() {
             if param.len() != 1 {
                 param.try_push_str(", ")?;
--- a/crates/rune/src/doc/autocomplete.rs
+++ b/crates/rune/src/doc/autocomplete.rs
@@ -1,0 +1,394 @@
+use core::fmt::Display;
+
+use crate::doc::{Artifacts, Context, Visitor};
+
+use anyhow::{Context as _, Result};
+use pulldown_cmark::{Options, Parser};
+use rune_alloc::{fmt::TryWrite, HashMap, String};
+use rune_core::{Hash, ItemBuf};
+use syntect::parsing::SyntaxSet;
+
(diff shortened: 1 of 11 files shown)


udoprog:
Very neat! Could you include an example how this is used in the PR description?

crates/rune/src/doc/autocomplete.rs
```diff
@@ -0,0 +1,388 @@
...
+        markdown::push_html(&self.syntax_set, &mut o, iter, None)?;
+
+        write!(o, "</div>")?;
+        let o = String::try_from(o.replace('`', "\\`"))?;
+
+        Ok(Some(o))
+    }
+
+    fn get_name(item: &ItemBuf) -> Result<String> {
+        let mut name = item.try_to_string()?;
```
udoprog, line 143: Some of the below seems to be using the wrong range, for example `::std::io::println!()` is longer than `12`, but I'm also curious: What is this doing? :sweat_smile:

A better implementation would probably be:

```rust
const PATTERNS: &[&str] = &["foo", "bar"];

for pat in PATTERNS {
    if name.starts_with(pat) {
        name.try_replace_range(..pat.len(), "")?;
    }
}

Ok(name)
```

udoprog: All right, so the build includes a [prefix](https://github.com/rune-rs/rune/blob/main/crates/rune/src/build.rs#L280). It might be possible to make this available somehow. It includes all the things imported by default.

Roba1993: Ah perfect! That's exactly what I looked for. Now it works as expected.
From my perspective this PR is now ready to go.

Roba1993:
Is there anything still to do, so that we can merge this PR?

udoprog:
I don't think there's any outstanding issues, I'll rebase it once I'm done with the `fmt` refactoring.

udoprog:
I've rebased this and updated the PR to work with the latest upstream. I'm working on including a local example so I can try it out, then I'll merge it.

PR #715, opened 2024-06-20
Convert Function::call (and related) to take GuardedArgs instead of Args

dstoza:


crates/rune/src/runtime/function.rs
```diff
@@ -520,21 +521,17 @@ where
             }
             Inner::FnTupleStruct(tuple) => {
                 vm_try!(check_args(args.count(), tuple.args));
-                vm_try!(Value::tuple_struct(
-                    tuple.rtti.clone(),
-                    vm_try!(args.try_into_vec())
-                ))
+                let (vec, _guard) = vm_try!(unsafe { args.unsafe_into_vec() });
```
udoprog, line 524: Could you add test cases for these?

What should happen if a reference is used is that the value tuple returned contains values which return `false` when checked with [`Value::is_readable` and `Value::is_writable`](https://github.com/rune-rs/rune/blob/999eac032df1148baadde61fc412bebd23630b12/crates/rune/src/runtime/value.rs#L555).

udoprog (approved):
Looks correct to me with one comment, thank you!

crates/rune/src/runtime/function.rs
dstoza: Just a quick note that I do intend to do this at some point. I'm just really busy with summer travel and other work, so it may be a few weeks.

dstoza: I think the tests I added are what you were asking for, but let me know if I misunderstood. Sorry for the delay 🙃

udoprog: So I'm pretty sure this never works, because the guard cannot possibly ever be held for the duration of which the tuple you constructed is used.

I don't think it's unsound though. It just means the values contained in the tuple will never be accessible. Writing a test case which uses this call function and tries to use the interior value somehow should reveal this.

dstoza: Ah, I think I see what you're saying now. In that case, should call fail for FnTuple{Struct,Variant} if there are references, or do we need some other signature that excludes them for GuardedArgs?

.github/workflows/ci.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       matrix:
         include:
           - crate: rune
-            target: "runtime:: hir::arena:: --skip runtime::vm:: --skip runtime::vm_execution::"
+            target: "runtime:: hir::arena:: --skip runtime::vm:: --skip runtime::vm_execution:: --skip runtime::function::"
```
dstoza, line 80: This does technically skip one existing test, `assert_send_sync`, but it seems like that test is designed to fail at (test) compile time, so it shouldn't need to be run under miri

dstoza:
Okay, now that I actually understand your point, I think I've fixed this correctly: `Function::call` still takes `GuardedArgs`, but rather than uselessly returning TupleVariant/TupleStructs with internal references that can't (and shouldn't) be accessed, it returns an error.

.github/workflows/ci.yml
udoprog: Why are we skipping over the `runtime::function` tests to begin with?

crates/rune/src/runtime/function.rs
```diff
@@ -1058,4 +1064,101 @@ mod tests {
         assert_send::<SyncFunction>();
         assert_sync::<SyncFunction>();
     }
+
+    #[derive(Default)]
+    struct MyAny {}
```
udoprog, line 1069: Please move these tests into a dedicated file in the [`tests` folder](https://github.com/rune-rs/rune/tree/main/crates/rune/src/tests).

.github/workflows/ci.yml
dstoza: The tests I added there seemed to be taking a very long time to run under miri CI, and I didn't want to clog up the CI build for everyone else. It shouldn't be a problem after I move them to the `tests` folder, though.

crates/rune/src/runtime/function.rs
dstoza: Done, thanks!

udoprog:
Cool, thank you!

PR #723, opened 2024-07-06
modules: add base64 module

Roba1993:
This adds basic base64 encoding & decoding


crates/rune-modules/src/base64.rs
```diff
@@ -23,7 +23,7 @@ pub fn module(_stdio: bool) -> Result<Module, ContextError> {
 /// ```
 #[rune::function]
 fn decode(inp: &str) -> Result<Vec<u8>, String> {
-    Ok(BASE64_STANDARD.decode(inp).map_err(|e| e.to_string())?)
```
udoprog: Please define a wrapper around the base64 error typ8e, else the error type exported to Rune scripts will be strings rather than an appropriate error type.

The error type primarily needs to implement the `STRING_DISPLAY` protocol. You can see [an example here](https://github.com/rune-rs/rune/blob/52c52f5d93c2336aa4434cd713c619384b70a2c1/crates/rune-modules/src/http.rs#L96).

crates/rune-modules/src/base64.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    module.function_meta(decode)?;
+    module.function_meta(encode)?;
+    Ok(module)
+}
+
+/// Decode a base64 String into data
+///
+/// ```rune
+/// //assert_eq!(base64::decode("+uwgVQA=")?, b"\xFA\xEC\x20\x55\0");
```
udoprog, line 21: Is this test intentionally commented out? If it's intended to fail, maybe do something like this:

```
assert!(base64::decode("+uwgVQA=").is_err());
```

crates/rune-modules/src/base64.rs
udoprog: Moving this conversation.

crates/rune-modules/src/base64.rs
```diff
@@ -0,0 +1,37 @@
...
+/// assert!(base64::decode("+uwgVQA=").is_ok());
+/// ```
+#[rune::function]
+fn decode(inp: &str) -> Result<Vec<u8>, String> {
+    BASE64_STANDARD.decode(inp).map_err(|e| e.to_string())
```
udoprog, line 26: Please define a wrapper around the base64 error typ8e, else the error type exported to Rune scripts will be strings rather than an appropriate error type.

The error type primarily needs to implement the `STRING_DISPLAY` protocol. You can see [an example here](https://github.com/rune-rs/rune/blob/52c52f5d93c2336aa4434cd713c619384b70a2c1/crates/rune-modules/src/http.rs#L96).

crates/rune-modules/src/base64.rs
Roba1993: Yeah good point. Will do it

crates/rune-modules/src/base64.rs
Roba1993: I commented it out, because I initially implemented it for an older version of rune, where comparing bytes was not possible. It should work now?

crates/rune-modules/src/base64.rs
Roba1993: Done, Thanks for the hint.

crates/rune-modules/src/base64.rs
Roba1993: Not working now, I need to find the location to support to compare two `Vec<u8>` within rune...

Roba1993: I will return rune Bytes which is most likely the more correct approach.

crates/rune-modules/src/base64.rs
```diff
@@ -0,0 +1,64 @@
...
+#[rune::function]
+fn encode(bytes: &[u8]) -> String {
+    BASE64_STANDARD.encode(bytes)
+}
+
+/// An error returned by methods in the `base64` module.
+#[derive(Debug, rune::Any)]
+#[rune(item = ::base64)]
+#[allow(dead_code)]
+pub struct Error {
```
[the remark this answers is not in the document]
Roba1993: Yes, done,

crates/rune-modules/src/base64.rs
```diff
@@ -0,0 +1,64 @@
...
+    Ok(BASE64_STANDARD.decode(inp)?)
+}
+
+/// Encode a data into a base64 String.
+///
+/// ```rune
+/// assert_eq!(base64::encode(b"\xFF\xEC\x20\x55\0"), "/+wgVQA=");
+/// ```
+#[rune::function]
+fn encode(bytes: &[u8]) -> String {
```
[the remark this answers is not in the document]
Roba1993: Yes, that's a good idea. Done

crates/rune-modules/src/base64.rs
```diff
@@ -0,0 +1,64 @@
...
+    Ok(module)
+}
+
+/// Decode a base64 String into data
+///
+/// ```rune
+/// assert_eq!(base64::decode("+uwgVQA=")?, b"\xFA\xEC\x20\x55\0");
+/// ```
+#[rune::function]
+fn decode(inp: &str) -> Result<Vec<u8>, Error> {
```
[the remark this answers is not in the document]
Roba1993: I'm using rune::runtime::Bytes now to fulfill memory limit requirements and support a correct comparison.

Roba1993:
Should be ready for merge now

crates/rune-modules/src/base64.rs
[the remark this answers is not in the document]
udoprog: So the method still goes through an intermediate `std::string::String`, which perform an allocation that is outside of Rune's purview.

You'll have to use [`encode_slice`](https://docs.rs/base64/latest/base64/engine/trait.Engine.html#method.encode_slice) writing into a buffer of size [`encode_len`](https://docs.rs/base64/latest/base64/fn.encoded_len.html) to fully avoid this.

You can see how it's used in the [default `encode` implementation](https://docs.rs/base64/latest/src/base64/engine/mod.rs.html#120).

Issue #747, opened 2024-07-16
Panic while formatting anyhow `Result` wrapped in rune `Value`


VorpalBlade:
Given the following example code (full reproducer project attached as tar.gz below):

```rust
use std::sync::Arc;

use anyhow::Context;
use rune::{
    termcolor::{ColorChoice, StandardStream},
    ContextError, Diagnostics, Module, Source, Sources, Vm,
};

/// Simple rune extension function that returns anhow error
#[rune::function]
fn hello() -> anyhow::Result<()> {
    Err(anyhow::anyhow!("AIEEEEE")).context("Blah")?;
    Ok(())
}

// Module
#[rune::module(::dummy)]
pub(crate) fn module() -> Result<Module, ContextError> {
    let mut m = Module::from_meta(self::module_meta)?;
    m.function_meta(hello)?;
    Ok(m)
}

#[tokio::main(flavor = "current_thread")]
async fn main() -> anyhow::Result<()> {
    // Set up logging with tracing
    let filter = tracing_subscriber::EnvFilter::builder()
        .with_default_directive(tracing::level_filters::LevelFilter::INFO.into())
        .from_env()?;
    let subscriber = tracing_subscriber::fmt::Subscriber::builder()
        .with_env_filter(filter)
        .finish();
    tracing::subscriber::set_global_default(subscriber)?;

    // Create rune context
    let mut context = rune::Context::default();
    context.install(module()?)?;

    let runtime = Arc::new(context.runtime()?);

    // Load example code
    let code = r#"
    pub fn main() {
        ::dummy::hello()?;
    }"#;
    let mut sources = Sources::new();
    sources.insert(Source::memory(code)?)?;

    // Diagnostics
    let mut diagnostics = Diagnostics::new();

    let result = rune::prepare(&mut sources)
        .with_context(&context)
        .with_diagnostics(&mut diagnostics)
        .build();

    if !diagnostics.is_empty() {
        let mut writer = StandardStream::stderr(ColorChoice::Always);
        diagnostics.emit(&mut writer, &sources)?;
    }
    let unit = result?;
    let mut vm = Vm::new(runtime, Arc::new(unit));

    // Do an async call
    let vm_result = vm.async_call(["main"], ()).await;

    // Print VM-result:
    dbg!(&vm_result);
    Ok(())
}
```

we get the following panic while running it:

```
[...]

PR #748, opened 2024-07-17
rune: Do not trace return value (causes panic on Result:Err)

VorpalBlade:
This is the fix for  #747 on the main branch (removal of instrumenting the return value). I left the instrument itself in, though this could be removed instead if it isn't useful to keep that around.

### DIFF
--- a/crates/rune/src/runtime/vm.rs
+++ b/crates/rune/src/runtime/vm.rs
@@ -2239,7 +2239,7 @@ impl Vm {
     }
 
     #[inline]
-    #[tracing::instrument(skip(self))]
+    #[tracing::instrument(skip(self, return_value))]
     fn op_return_internal(&mut self, return_value: Value) -> VmResult<Option<Output>> {
         let (exit, out) = vm_try!(self.pop_call_frame());
 


udoprog:
Thanks!

Wouldn't fixing the `fmt::Debug` impl make this unnecessary?

VorpalBlade:
Hm, fair point, I work primarily on 0.13.x, not main. So that variant is more well tested. However I don't think it would currently crash on main. But if you somehow manage to return a moved value it would. Not sure if that is possible.

VorpalBlade:
Do you want to remove the tracing change now? It *shouldn't* crash without that any more, but I admit to not having tested that yet. Even if it doesn't crash, how does every return attempting to format the value affect the performance?

udoprog:
> Do you want to remove the tracing change now? It *shouldn't* crash without that any more, but I admit to not having tested that yet. Even if it doesn't crash, how does every return attempting to format the value affect the performance?

It should only affect performance if tracing is enabled. Otherwise it should be negligible. Tracing can also be statically compiled away with feature flags if we want to maximize performance.

crates/rune/src/runtime/value.rs
```diff
@@ -2137,7 +2137,15 @@ impl fmt::Debug for Value {
         let mut o = Formatter::new();
 
         if self.string_debug(&mut o).is_err() {
```
udoprog, line 2139: Might be useful to include this error as well?

udoprog (approved):

PR #756, opened 2024-07-21
docs: Add missing docs for module(keep)

### LINKED ISSUE #754: Missing docs for `#[rune::function(keep)]`
I have been annoyed for a while that I can't call the Rune extension methods from Rust (when I need them from both locations). I was considering filing a feature request on this, but while looking through rune's own HashMap there is a `keep` attribute that seems to solve this nicely already.

But it is not documented in https://rune-rs.github.io/api/rune/attr.function.html

So: Is the docs missing (in which case I will make a PR) or is it an unstable private API (in which case I propose to stabilise it and document it)?

VorpalBlade:
Fixes #754

I have not made any fix for 0.13. Don't know if we want to backport it (should be very simple if so)

### DIFF
--- a/crates/rune/src/lib.rs
+++ b/crates/rune/src/lib.rs
@@ -581,7 +581,7 @@ pub(crate) use rune_macros::__internal_impl_any;
 ///
 /// By default, the name of the function is mangled and the metadata is given
 /// the original name. This means you can't easily call the function from both
-/// Rune and Rust. This behaviour can be changed by using the `keep` option, in
+/// Rune and Rust. This behaviour can be changed by using the `keep` attribute, in
 /// which case you must refer to the meta object by a mangled name
 /// (specifically the function name with `__meta` appended):
 ///
--- a/crates/rune/src/lib.rs
+++ b/crates/rune/src/lib.rs
@@ -577,6 +577,34 @@ pub(crate) use rune_macros::__internal_impl_any;
 /// }
 /// ```
 ///
+/// # Using `keep` to keep the name
+///
+/// By default, the name of the function is mangled and the metadata is given
+/// the original name. This means you can't easily call the function from both
+/// Rune and Rust. This behaviour can be changed by using the `keep` option, in
+/// which case you must  refer to the meta object by a managled name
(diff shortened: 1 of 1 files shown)


crates/rune/src/lib.rs
```diff
@@ -577,6 +577,34 @@ pub(crate) use rune_macros::__internal_impl_any;
 /// }
 /// ```
 ///
+/// # Using `keep` to keep the name
+///
+/// By default, the name of the function is mangled and the metadata is given
+/// the original name. This means you can't easily call the function from both
+/// Rune and Rust. This behaviour can be changed by using the `keep` option, in
+/// which case you must  refer to the meta object by a managled name
```
udoprog, line 585:
```suggestion
/// which case you must refer to the meta object by a mangled name
```

crates/rune/src/lib.rs
```diff
@@ -577,6 +577,34 @@ pub(crate) use rune_macros::__internal_impl_any;
...
+/// (specifically the function name with `__meta` appended).:
```
udoprog, line 586:
```suggestion
/// (specifically the function name with `__meta` appended):
```

crates/rune/src/lib.rs
```diff
@@ -577,6 +577,34 @@ pub(crate) use rune_macros::__internal_impl_any;
 /// }
 /// ```
 ///
+/// # Using `keep` to keep the name
+///
+/// By default, the name of the function is mangled and the metadata is given
+/// the original name. This means you can't easily call the function from both
+/// Rune and Rust. This behaviour can be changed by using the `keep` option, in
```
udoprog, line 584:
```suggestion
/// Rune and Rust. This behaviour can be changed by using the `keep` attribute, in
```

udoprog (changes requested):

VorpalBlade:
Merged your changes (on the phone at the moment, and since you apparently squash merge anyway it doesn't matter that I normally try to make clean mergable histories)

udoprog (approved):

udoprog:
Thanks!

Issue #633, opened 2023-09-24
request: support traits


Fluffyalien1422:
There are already `impl` blocks but no traits

Traits can be syntax sugar for regular `impl` blocks

```rs
trait Y {
	fn say_hi() {
		println!("Hi!");
	}
}

struct X {}
impl Y for X {}
```

is the same as:

```rs
struct X {}

impl X {
	fn say_hi() {
		println!("Hi!");
	}
}
```

Fluffyalien1422:
I've started working on this here https://github.com/Fluffyalien1422/rune

udoprog:
Also see #23

Some thoughts:

Since traits and their associated methods can be defined elsewhere (e.g. other crates where they trait is defined), they can't strictly be syntactic sugar for impl blocks. So how do you propose dealing with method naming conflicts and potential disambiguation?

How would method name resolution work? In Rust, the methods of a trait are only visible implicitly if the trait is in scope, but this relies on knowing the types of values which we obviously don't in Rune.

In #23 traits are solely used to define protocols. And this could be a nice area for you to explore since we currently can't implement protocols in pure Rune.

Thanks!

Fluffyalien1422:
> How would method name resolution work? In Rust, the methods of a trait are only visible implicitly if the trait is in scope, but > this relies on knowing the types of values which we obviously don't in Rune.

idk how the codebase works exactly yet so idk if this will actually work but here's the idea:
A trait would really only exist for compiler errors and default implementations, so it won't matter if the trait is in scope to use the methods or not because implementing a trait will be the same as just a normal `impl` block
So
```rs
trait X {
	fn no_default_implementation();
	fn default_implementation() {
		println!("Hello, World!");
	}
}

struct Y {};
impl X for Y {
	fn no_default_implementation() {
		println!("Bye!");
	}
}
```
Is the same as
```rs
struct Y {};
impl Y {
	fn no_default_implementation() {
		println!("Bye!");
	}
	fn default_implementation() {
		println!("Hello, World!");
	}
}
```
So it wouldn't matter if the trait is in scope because the trait only exists to give compiler errors and to fill in default implementations, it will actually just end up being exactly the same as a normal impl block

> Since traits and their associated methods can be defined elsewhere (e.g. other crates where they trait is defined), they can't
> strictly be syntactic sugar for impl blocks. So how do you propose dealing with method naming conflicts and potential
> disambiguation?

```rs
struct X {}
impl a::b::C for X {}
```
[...]

PR #774, opened 2024-07-24
First iteration at implementing traits

udoprog:
This documents how the first implementation of traits is done in Rune.

Previously there were no such thing as a "trait", a type could opt to implement protocols which on a per-function basis implements trait-like behavior. Such as converting something into an iterator.

These protocols are what's referenced by the virtual machine. `Protocol::INTO_ITER` is for examples used in loop expressions such as `for <pat> in <expr>`.

This introduces traits as a concept in Rune.

#### Introduction

Traits are a collection of named associated functions. Once a trait is implemented, what that implies is simply that we can be sure that all the associated names it defines are present on the type.

As usual, Rune doesn't permit more than one definition of an associated name. Attempting to define more than one results in a build-time error.

In the background the implementation of traits is based on protocols just before. For example, in order to implement the `Iterator` trait, a type has to implement the `ITER` protocol.

This separation is important because we don't want a user to *accidentally* implement an associated method which would then be legally used in a trait. Protocols are defined in their own namespace, and cannot be invoked in user code.

Instead, if the `ITER` protocol is present and we request that the `::std::iter::Iterator` trait should be implemented, the `ITER` protocol implementation is used to construct all the relevant associated methods. This is done by calling `Module::implement_trait`.

```rust
let mut m = Module::new();
m.ty::<Iter>()?;
m.function_meta(Iter::next__meta)?;
m.function_meta(Iter::size_hint__meta)?;
m.function_meta(Iter::nth__meta)?;
m.function_meta(Iter::next_back__meta)?;
m.implement_trait::<Iter>(rune::item!(::std::iter::Iterator))?;
m.implement_trait::<Iter>(rune::item!(::std::iter::DoubleEndedIterator))?;
```

Note that this allows the `Iter` type above to specialize its `NTH` implementation, but for example the `nth_back` one will use a default one.

The result is that the `Iter` type *automatically* gets all the iterator-associated methods added to it.

#### Defining a trait

Currently defining a trait is a low-level business, but technically it's done by implementing a handler which will be called to populate the relevant methods. Such as this snipped for the `Iterator` trait:

```rust
let mut m = Module::new();

let mut t = m.define_trait(["Iterator"])?;

t.handler(|cx| {
    let next = cx.find(Protocol::NEXT)?;
    cx.function_handler("next", &next)?;

    let size_hint = if let Some(size_hint) = cx.try_find(Protocol::SIZE_HINT)? {
        cx.function_handler("size_hint", &size_hint)?;
        size_hint
    } else {
        let size_hint = cx.function("size_hint", |_: Value| (0usize, None::<usize>))?;
        cx.function_handler(Protocol::SIZE_HINT, &size_hint)?;
        size_hint
    };

    /* more methods */
    Ok(())
})?;
```

Here we can see that the implementation for `SIZE_HINT` will fallback to a default implementation if it's not implemented, and the appropriate protocol is also populated if it's missing. Meanwhile `NEXT` is required to be implemented.


udoprog:
Relates #633

Issue #782, opened 2024-07-28
Formatting likes to add more and more blank lines (not idempotent)


VorpalBlade:
Given:

```rust
pub fn test() {

    let ctx = Context {
        props,
        cmds,
        package_managers,
        system,
        features,
        tasks,
        passwd,
    };
}
```

rune's autoformatting likes to insert one extra blank line every time it is executed, after a few runs you get something like:

```rust
pub fn test() {

    let ctx = Context {
        props,
        cmds,
        package_managers,
        system,
        features,
        tasks,
        passwd,




    };
}
```

I have also observed this between top level constants:

```rust
pub const USER_MAPPING = [
    ("qemu", 954),
    ("saned", 956),
    ("passim", 957),
    ("_talkd", 958),
    ("fwupd", 960),
    ("systemd-journal-upload", 962),
    ("sddm", 963),
    ("icecream", 965),
    ("distcc", 966),
    ("tss", 968),
    ("flatpak", 969),
    ("geoclue", 970),
    ("colord", 971),
    ("avahi", 972),
    ("chrony", 973),
    ("dnsmasq", 974),
    ("git", 975),
    ("systemd-timesync", 976),
    ("systemd-resolve", 977),
    ("systemd-journal-remote", 978),
    ("systemd-oom", 979),
    ("systemd-network", 980),
    ("systemd-coredump", 981),
];





pub const GROUP_MAPPING = [
    ("qemu", 954),
    ("gamemode", 955),
    ("saned", 956),
    ("passim", 957),
    ("_talkd", 958),
    ("groups", 959),
    ("fwupd", 960),
    ("cdemu", 961),
    ("systemd-journal-upload", 962),
    ("sddm", 963),
    ("i2c", 964),
    ("icecream", 965),
    ("distcc", 966),
    ("adbusers", 967),
    ("tss", 968),
    ("flatpak", 969),
    ("geoclue", 970),
    ("colord", 971),
    ("avahi", 972),
    ("chrony", 973),
    ("dnsmasq", 974),
    ("git", 975),
    ("systemd-timesync", 976),
    ("systemd-resolve", 977),
    ("systemd-journal-remote", 978),
    ("systemd-oom", 979),
    ("systemd-network", 980),
    ("systemd-coredump", 981),
    ("rfkill", 982),
    ("systemd-journal", 983),

    ("users", 984),
    ("video", 985),
    ("uucp", 986),
    ("storage", 987),
    ("sgx", 988),
    ("render", 989),
[...]

PR #790, opened 2024-07-29
fmt: First stab at lexer-based formatter

udoprog:
This is the first stab at rewriting the formatter to be lexer-based rather than parser based.

While the parser-based format is better, there are several key limitations I'd like to address:

Be able to format code which is syntactically incorrect, such as having or not having terminating commas or semi-colons or missing blocks:

```rust
struct     Foo { foo,,, }
struct     Foo  (foo,,,)
struct    Foo
struct    Foo    ;;;
enum Foo {Bar}
enum Foo
```

Should format to:

```rust
struct Foo {
    foo,
}
struct Foo(foo);
struct Foo;
struct Foo;
enum Foo {
    Bar,
}
enum Foo {}
```

Have a reasonably stable foundation where formatting can be refined as-needed. This is achieved by losslessly parsing the tree and if you don't want to format a specific component you can simply reference a parent node and get the literal content.

The compiler should tell you if you forget to write something. With the new architecture this is achieved by marking the relevant syntax nodes with `#[must_use]`, which gives you a compiler warning if you haven't consumed them. Consuming them means you've asserted that they have been written to the output.

#### New parser and grammar

With this comes a new parser which uses a lossless grammar to parse Rune source. This is intended to be incorporated into the wider project, it is particularly interesting to use during indexing, because partially incorrect source can still be indexed and usable by the language server.

The grammar has so far been designed to be as permissive as possible without causing parsing ambiguities. That means that statements like these are grammatically parseable, but it's obviously not valid Rune:

```rust
let values = [1, 2, 3

enum Bar {
   Foo,
```

Here the parser will "ignore" that the closing bracket and brace is missing. The bracket can be ignored because the succeeding token `stmt` is neither a valid expression, nor a valid terminator for the array (which is `]`).

Coincidentally, formatting the above will currently add the missing terminating tokens. This behavior might however be changed if it's broadly disliked.

#### Line hints

The input source is used to extract "line hints", so with a source like this:

```rust
let foo = 42;

let bar = 42;

1 + 2
```

At most two line breaks would be preserved between the statements, however, given something like this:

```rust
let foo = 42; let bar = 42; 1 + 2
```

We would get:

```rust
let foo = 42;
let bar = 42;

1 + 2
```

That's because the previous block had embedded lines which we used as hints when formatting. While for the second one, the formatting engine gets to decide the defaults instead.

#### `#[runefmt::skip]`

This attribute is an escape hatch that can be used to disable formatting entirely for an item or an expression.

For example:

```rust
#[runefmt::skip]
let var = [
    1, 2,
    3, 4,
    5, 6,
    7, 8,
    9, 10
];

let var = [
    1, 2,
    3, 4,
    5,
];

#[runefmt::skip]
let var = [
    1, 2,
    3, 4,
    5, 6,
    7, 8,
    9, 10
];
```


udoprog:
@VorpalBlade would you mind taking this for a spin? I'm hoping it should address formatting bugs like #782.

VorpalBlade:
It throws a ton of errors when I attempt to use the rune cli for this. Probably because I'm using the generic one and not one with my modules added (my code targets 0.13.x and I haven't looked at how much porting work it would be for 0.14.x). Lots of "missing item". Apart from one there is also:

```
error: 
   ┌─ /home/arvid/src/syskonf/tasks/grub.rn:33:11
   │
33 │         } else {
   │           ^^^^ Unsupported expression Else
```

The code in question:

```rust
    if ctx.features.kdump {
        if ctx.system.hw_memory < 8 * 1024 * 1024 {
            cmd_line_default.push("crashkernel=128M");
        } else {
            cmd_line_default.push("crashkernel=256M");
        }
    }
```

So that is a bit odd?

That said, some good news:

* The code as formatted is still valid rune.
* Note more spurious blank lines.
* The weirdness with un-indented comments right before the end of a block is gone.

Not all good news though:

## Issue 1

This still ends up on one line:

```rust
    let edit_commands = [
        (Selector::Regex("^HOOKS="), Action::RegexReplace("=\\((.*)\\)$", std::fmt::format!("=({})", crate::utils::join(' ', hooks)))),
        (Selector::Regex("^MODULES="), Action::RegexReplace("=\\((.*)\\)$", std::fmt::format!("=({})", crate::utils::join(' ', modules)))),
    ];
```

becomes

```rust
    let edit_commands = [(Selector::Regex("^HOOKS="), Action::RegexReplace("=\\((.*)\\)$", std::fmt::format!("=({})", crate::utils::join(' ', hooks)))), (Selector::Regex("^MODULES="), Action::RegexReplace("=\\((.*)\\)$", std::fmt::format!("=({})", crate::utils::join(' ', modules))))];
```

(I personally think about 100 charaters should be the maximum line length aimed for)

## Issue 2

This is very strange, and a variation of the above one:

```rust
    patch_file(ctx, "bash", "/etc/skel/.bashrc", [
        (Selector::Regex("force_color_prompt"), Action::RegexReplace("^( *)#", "$1")),
        (Selector::Regex("PS1=.*01;32m"), Action::Replace("    PS1='${debian_chroot:+($debian_chroot)}\\[\\033[01;32m\\]\\u@\\h\\[\\033[00m\\]:\\[\\033[01;34m\\]\\w \\$\\[\\033[00m\\] '")),
        (Selector::Regex("alias (grep|fgrep|egrep)"), Action::RegexReplace("^( *)#", "$1")),
    ])?;

    patch_file(ctx, "passwd", "/etc/default/useradd", [
        (Selector::All, Action::RegexReplace("SHELL=/bin/sh", "SHELL=/bin/bash")),
        (Selector::Regex("SKEL="), Action::RegexReplace("^# *", "")),
    ])?;

    patch_file_to(ctx, "libc-bin", "/usr/share/libc-bin/nsswitch.conf", "/etc/nsswitch.conf", [
        (Selector::Regex("^hosts:"), Action::RegexReplace("files dns$", "files mdns4_minimal [NOTFOUND=return] dns"))
    ])?;
    patch_file(ctx, "login", "/etc/login.defs", [
        (Selector::Regex("ENV_PATH"), Action::RegexReplaceAll("(/usr[^:]*/)(bin):", "${1}s$2:$1$2:")),
        (Selector::Regex("ENV_PATH"), Action::RegexReplaceAll(":/bin:", ":/sbin:/bin:"))
    ])?;

    patch_file_to(ctx, "base-files", "/usr/share/base-files/profile", "/etc/profile", [
        (Selector::Regex("PATH=\"/usr/local/bin"), Action::RegexReplaceAll("(/usr[^:]*/)(bin):", "${1}s$2:$1$2:")),
        (Selector::Regex("PATH=\"/usr/local/bin"), Action::RegexReplaceAll(":/bin:", ":/sbin:/bin:"))
    ])?;
    ctx.cmds.write("/etc/motd", b"")?;
```

Gets turned into:

```rust
    patch_file(ctx, "bash", "/etc/skel/.bashrc", [(Selector::Regex("force_color_prompt"), Action::RegexReplace("^( *)#", "$1")), (Selector::Regex("PS1=.*01;32m"), Action::Replace("    PS1='${debian_chroot:+($debian_chroot)}\\[\\033[01;32m\\]\\u@\\h\\[\\033[00m\\]:\\[\\033[01;34m\\]\\w \\$\\[\\033[00m\\] '")), (Selector::Regex("alias (grep|fgrep|egrep)"), Action::RegexReplace("^( *)#", "$1"))])?;
    patch_file(ctx, "passwd", "/etc/default/useradd", [(Selector::All, Action::RegexReplace("SHELL=/bin/sh", "SHELL=/bin/bash")), (Selector::Regex("SKEL="), Action::RegexReplace("^# *", ""))])?;
    patch_file_to(ctx, "libc-bin", "/usr/share/libc-bin/nsswitch.conf", "/etc/nsswitch.conf", [(Selector::Regex("^hosts:"), Action::RegexReplace("files dns$", "files mdns4_minimal [NOTFOUND=return] dns"))])?;
    patch_file(ctx, "login", "/etc/login.defs", [(Selector::Regex("ENV_PATH"), Action::RegexReplaceAll("(/usr[^:]*/)(bin):", "${1}s$2:$1$2:")), (Selector::Regex("ENV_PATH"), Action::RegexReplaceAll(":/bin:", ":/sbin:/bin:"))])?;
    patch_file_to(ctx, "base-files", "/usr/share/base-files/profile", "/etc/profile", [(Selector::Regex("PATH=\"/usr/local/bin"), Action::RegexReplaceAll("(/usr[^:]*/)(bin):", "${1}s$2:$1$2:")), (Selector::Regex("PATH=\"/usr/local/bin"), Action::RegexReplaceAll(":/bin:", ":/sbin:/bin:"))])?;
    ctx.cmds.write("/etc/motd", b"")?;
```

This is not *just* the previous issue. Somehow the blank lines in between got lost too. In fact it seems every time I have a multi-line patch_file-style call it also swallows all newlines afterwards when it reformats.

## Issue 3

```rust
impl System {
    pub fn new(sysinfo /* sysinfo::SysInfo */) {
        // ...
    }
}
```

becomes 

```rust
impl System {
    pub fn new(sysinfo) /* sysinfo::SysInfo */ {
        // ...
    }
}
```

Not sure what is going on there, and I don't know what it would do with a multi-argument function (e.g. `f(a /* A */, b /* B */)`) as I don't have one of those in my code base.

udoprog:
Thank you!

Issue 2 will be solved once reflowing is implemented. we could theoretically do that now, but we'd have to base reflowing decisions on the raw input rather than the desired output which might make formatting unstable. Instead I intend to build another linebreak syntax tree based on the existing hints which will handle reflowing based on the size of the desired output correctly.

Issue 3 is something I need to think over. I've intentionally had it gobble up comments right now looking for checkpoints where they can be inserted, and expressions like `(arg)` (without whitespace) currently doesn't have a natural checkpoint. There are no whitespace or line hints to work with.

udoprog:
So I've added raw heuristics as a stopgap measure for Issue 2. Please take it for another spin!

Also note that you don't have to build your project to run it, you can check out rune in a separate repo and run the formatter like this:

```
cargo run --bin rune -- fmt --recursive --check <path>
```

VorpalBlade:
I will give this another spin tomorrow, after which I will be unavailable for about a week.

VorpalBlade:
This latest version is much better. I would be happy with it as is. Just some *minor* nitpicks:

## Nitpick 1

```rust
        patches.push(
            (Selector::Regex("# begin section managed by etckeeper"), Action::InsertAfter(
                "\n# new and old versions of conffiles, stored by pacman\n*.pacnew\n*.pacorig\n*.pacsave",
            )),
        );
```

I think it might make more sense to try to break the line at a higher level of nesting (just before `Action::`), but perhaps the line was too long that way, I don't know. It is a minor nitpick at most.

## Nitpick 2

```rust
                ctx
                    .cmds
                    .copy_from("/etc/firewalld/zones/home.xml", `/etc/firewalld/zones/home.${hostname}.xml`)?;
```

I'm not sure `.cmds` needs to be on it's own line here, but in *general* I do think it is better to go for either "everything multi-line in the call chain" or "everything on line line" (just like rustfmt) does. I think this is just a silly degenerate case perhaps. Another minor nitpick. Clearly it is better in this case for example (which I think is more typical):

```rust
    let net_type = lines
        .iter()
        .find(|line| line.starts_with("type="))
        .map(|v| v.split("=").collect::<Vec>()[1]);
```

Perhaps the rule should be that there shouldn't be a linebreak after the first item in a call chain if it is shorter than the indentation level?

VorpalBlade:
Seems good to me now. Thanks a lot for this.

udoprog:
Nitpicks we can live with, as long as it's not super borked!

Nitpick 2 has already been improved upon, with current heuristics it will be a single line, and if the arguments are longer it will look like this:

```rust
ctx.cmds.copy_from(
    "/etc/firewalld1234/zones/home.xml",
    `/etc/firewalld1234/zones/home.${hostname}.xml`,
)?;
```

It works by looking for the first `ExprCall` now instead and considering whether the chain up to it is short enough.

Thanks for the feedback!

udoprog:
> Seems good to me now. Thanks a lot for this.

You got it and thanks again. I've wanted to rewrite the parser for a while now to be lossless. This ended up being a good entrypoint for it. Now it can gradually be introduced into the rest of the compiler.

VorpalBlade:
You should probably test it on other code bases than just mine though.

I do plan to make a stripped down version of my personal system config public to be an example for users of my software (when I do get users of it). It is not a high priority at the moment because:

1. I need to strip out all personal info, and be sure I got it all. Can't take it back once I publish it on the internet.
2. The APIs are in a high amount of flux currently while I'm working on sensible multi-distro support (I started with Arch and I'm working on adding Debian/Ubuntu, though quite far along, I have a working config for my Raspberry Pi running Debian now.)

PR #813, opened 2024-08-21
Support non-ASCII identifiers

lotem:
Support Unicode identifiers.

Refer to the similar language feature in Rust:
https://github.com/rust-lang/rfcs/pull/2457


lotem:
Hi John-John @udoprog,

The `non_ascii_idents` feature is now merged to stable Rust. It would be nice to have Unicode identifier names in Rune too.
So far the feature seems to work after a few changes. Please advice if there's more to do.

udoprog:
Hi!

This looks great, thanks for the PR!

Please move the test into `crates/rune/tests`. They will be picked up automatically there.

A cargo feature flag is not appropriate since it transiently changes the behavior of anyone using Rune if one dependency activates it. In my view we should just adopt this for 0.14. If we do think it's an apptopriate optional feature, add a an `Options` flag so that individual projects can decide themselves whether or not to activate it.

lotem:
Thanks for the suggestions. I'll update the PR later.

lotem:
I was worried about introducing a dependency for a rarely needed feature. Otherwise I don't think users will ever want to disable it since it's a compatible change. Also the performance report from `unicode-ident` seems to be less of a concern.

udoprog:
The dependency is not a problem so please go ahead.

udoprog:
Thank you!

PR #815, opened 2024-08-30
modules: Expand time module with Duration, Interval and Instant support

ramon-bernardo:
This PR significantly enhances the time module introducing...

### Duration, now public:
- Added methods for creating durations from milliseconds, microseconds, nanoseconds, and floating-point seconds.
- Provided constants for common time units (SECOND, MILLISECOND, MICROSECOND, NANOSECOND).
- Implemented methods to convert and manipulate durations.

### Interval:
- Introduced the Interval struct with methods for managing and resetting time intervals.
- Added the ability to create asynchronous intervals that can be used to trigger repeated events.

### Instant:
- Introduced the Instant struct to represent specific points in time.
- Added methods to calculate the duration between instants and to check elapsed time.

The examples provided were adapted from the [Tokio Time](https://docs.rs/tokio/latest/tokio/time/index.html) library.

### DIFF
--- a/crates/rune-modules/src/time.rs
+++ b/crates/rune-modules/src/time.rs
@@ -29,139 +29,198 @@
 //! }
 //! ```
 
-use rune::{
-    docstring,
-    runtime::{Mut, VmResult},
-    vm_panic, Any, ContextError, Module,
-};
+use core::cmp::Ordering;
+use core::hash::Hash;
+
+use rune::alloc::fmt::TryWrite;
+use rune::runtime::{Formatter, Hasher, Mut, VmResult};
+use rune::{docstring, item, vm_panic, Any, ContextError, Module, ToConstValue};
 
 const NANOS_PER_SEC: u32 = 1_000_000_000;
 
 /// Construct the `time` module.
 pub fn module(_stdio: bool) -> Result<Module, ContextError> {
-    let mut module = Module::with_crate("time")?;
+    let mut m = Module::with_crate("time")?;
 
-    module.ty::<Duration>()?;
-    module.ty::<Interval>()?;
-    module.ty::<Instant>()?;
+    m.function_meta(sleep)?;
+    m.function_meta(interval)?;
+    m.function_meta(interval_at)?;
 
-    module.function_meta(Duration::new__meta)?;
-    module.function_meta(Duration::from_secs__meta)?;
-    module.function_meta(Duration::from_millis__meta)?;
-    module.function_meta(Duration::from_micros__meta)?;
-    module.function_meta(Duration::from_nanos__meta)?;
-    module.function_meta(Duration::as_secs_f64__meta)?;
-    module.function_meta(Duration::from_secs_f64__meta)?;
+    m.ty::<Duration>()?;
 
-    /* TODO: Make Duration a ConstValue
-    module
-        .constant("SECOND", Duration::from_secs(1))
-        .build_associated::<Duration>()?
-        .docs(docstring! {
-            /// The duration of one second.
-            ///
-            /// # Examples
-            ///
-            /// ```rune,no_run
-            /// use time::Duration;
-            ///
-            /// let duration = Duration::SECOND;
-            /// ```
-        })?;
+    m.function_meta(Duration::new__meta)?;
+    m.function_meta(Duration::from_secs__meta)?;
+    m.function_meta(Duration::from_millis__meta)?;
+    m.function_meta(Duration::from_micros__meta)?;
+    m.function_meta(Duration::from_nanos__meta)?;
+    m.function_meta(Duration::is_zero__meta)?;
+    m.function_meta(Duration::as_secs__meta)?;
+    m.function_meta(Duration::subsec_millis__meta)?;
+    m.function_meta(Duration::subsec_micros__meta)?;
+    m.function_meta(Duration::subsec_nanos__meta)?;
+    m.function_meta(Duration::as_millis__meta)?;
+    m.function_meta(Duration::as_micros__meta)?;
+    m.function_meta(Duration::as_nanos__meta)?;
+    m.function_meta(Duration::as_secs_f64__meta)?;
+    m.function_meta(Duration::from_secs_f64__meta)?;
+    m.function_meta(Duration::add__meta)?;
+    m.function_meta(Duration::add_assign__meta)?;
+    m.function_meta(Duration::partial_eq__meta)?;
+    m.implement_trait::<Duration>(item!(::std::cmp::PartialEq))?;
+    m.function_meta(Duration::eq__meta)?;
+    m.implement_trait::<Duration>(item!(::std::cmp::Eq))?;
+    m.function_meta(Duration::partial_cmp__meta)?;
+    m.implement_trait::<Duration>(item!(::std::cmp::PartialOrd))?;
+    m.function_meta(Duration::cmp__meta)?;
+    m.implement_trait::<Duration>(item!(::std::cmp::Ord))?;
+    m.function_meta(Duration::hash__meta)?;
+    m.function_meta(Duration::string_debug__meta)?;
+    m.function_meta(Duration::clone__meta)?;
+    m.implement_trait::<Duration>(item!(::std::clone::Clone))?;
 
-    module
-        .constant("MILLISECOND", Duration::from_millis(1))
-        .build_associated::<Duration>()?
-        .docs(docstring! {
-            /// The duration of one millisecond.
-            ///
-            /// # Examples
-            ///
-            /// ```rune,no_run
-            /// use time::Duration;
-            ///
-            /// let duration = Duration::MILLISECOND;
-            /// ```
-        })?;
+    m.constant(
+        "SECOND",
+        Duration {
+            inner: tokio::time::Duration::from_secs(1),
+        },
+    )
+    .build_associated::<Duration>()?
+    .docs(docstring! {
+        /// The duration of one second.
+        ///
+        /// # Examples
+        ///
+        /// ```rune
+        /// use time::Duration;
+        ///
+        /// let duration = Duration::SECOND;
+        /// ```
+    })?;
 
-    module
-        .constant("MICROSECOND", Duration::from_micros(1))
-        .build_associated::<Duration>()?
-        .docs(docstring! {
-            /// The duration of one microsecond.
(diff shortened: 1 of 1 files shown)


crates/rune-modules/src/time.rs
```diff
@@ -31,51 +31,334 @@
...
+}
+
+// #[derive(Debug, Any)]
+// #[rune(item = ::time)]
+// pub struct Timeout<T>
+// where
+//     T: Future + rune::compile::Named + CoreTypeOf + MaybeTypeOf + 'static,
+// {
+//     inner: tokio::time::Timeout<T>,
+// }
```
ramon-bernardo, line 364: Here

ramon-bernardo (commented):
How could we implement `Timeout<T>` in Rune @udoprog? I initially tried, but due to the complexities with `T`, I decided to pause for now... I'll try again later...

```rust
#[rune::function]
async fn timeout(duration: Duration, future: F) -> Timeout<F>
{
    Timeout {
        inner: tokio::time::timeout(duration.inner, future).await,
    }
}
```
```rust
#[derive(Debug, Any)]
#[rune(item = ::time)]
pub struct Timeout<T>
{
    inner: tokio::time::Timeout<T>,
}
```

ramon-bernardo:
Another situation is the async `tick` function; I'm encountering some issues with the `Future` trait.

```rust
    module
        .function("tick", Interval::tick)
        .build_associated::<Interval>()?;

    pub async fn tick(&mut self) {
        self.inner.tick().await;
    }
```


```rust
the trait bound 

for<'a> fn(&'a mut time::Interval) -> impl std::future::Future<Output = ()> {
    time::Interval::tick
}: rune::function::Function<_, _>

is not satisfied the trait `rune::function::Function<_, _>` is not implemented for fn item 

for<'a> fn(&'a mut time::Interval) -> impl std::future::Future<Output = ()> {
    time::Interval::tick
}
```

crates/rune-modules/src/time.rs
```diff
@@ -31,51 +31,334 @@
...
+    /// Creates a new `Duration` from the specified number of whole seconds and
+    /// additional nanoseconds.
+    ///
+    /// If the number of nanoseconds is greater than 1 billion (the number of
+    /// nanoseconds in a second), then it will carry over into the seconds provided.
+    ///
+    /// # Panics
+    ///
+    /// This constructor will panic if the carry from the nanoseconds overflows
+    /// the seconds counter.
```
udoprog, line 148: Rune methods must not actually panic, like a Rust panic since that could allow the script host to panic at runtime.

This must be checked, and the panicking behavior must use `VmResult::panic`.

crates/rune-modules/src/time.rs
```diff
@@ -31,51 +31,334 @@
...
+fn module_instant(module: &mut Module) -> Result<(), ContextError> {
+    module.ty::<Instant>()?;
+
+    module.function_meta(Instant::now__meta)?;
+    module.function_meta(Instant::duration_since__meta)?;
+    module.function_meta(Instant::elapsed__meta)?;
+
+    Ok(())
+}
+
```
udoprog, line 327: Missing docs for type?

crates/rune-modules/src/time.rs
```diff
@@ -31,51 +31,334 @@
...
+        .build_associated::<Interval>()?;
+
+    module.function_meta(Interval::reset__meta)?;
+    module.function_meta(Interval::reset_immediately__meta)?;
+    module.function_meta(Interval::reset_after__meta)?;
+    module.function_meta(Interval::reset_at__meta)?;
+
+    Ok(())
+}
+
```
udoprog, line 285: Missing docs for type?

crates/rune-modules/src/time.rs
```diff
@@ -31,51 +31,334 @@
...
+    module.function_meta(Duration::from_millis__meta)?;
+    module.function_meta(Duration::from_micros__meta)?;
+    module.function_meta(Duration::from_nanos__meta)?;
+    module.function_meta(Duration::as_secs_f64__meta)?;
+    module.function_meta(Duration::from_secs_f64__meta)?;
+
+    module
+        .constant("SECOND", Duration::from_secs(1))
+        .build_associated::<Duration>()?
+        .docs(["The duration of one second."])?;
```
udoprog, line 102: Try using `rune::docstring!`.

udoprog:
I'm not sure about `Timeout` for now. But principally it should not be generic in Rune since all futures are of a single type `Future`.

For `tick`, you might try to use a rune-specific function which takes `Mut<Self>`?

crates/rune-modules/src/time.rs
ramon-bernardo: Ready, docs added.

crates/rune-modules/src/time.rs
ramon-bernardo: Changed.

crates/rune-modules/src/time.rs
ramon-bernardo: Vm Panics, makes more sense?

udoprog: It does, but the new implementation still panics. You'll have to either use a checked operation when constructing the duration or [re-implement the logic](https://doc.rust-lang.org/nightly/src/core/time.rs.html#211) which is used to ensure that it's never fed values which panic.

crates/rune-modules/src/time.rs
```diff
@@ -29,53 +29,671 @@
...
     let mut module = Module::with_crate("time")?;
-    module.ty::<Duration>()?;
-    module.function_meta(Duration::from_secs__meta)?;
-    module.function_meta(sleep)?;
+
+    module_duration(&mut module)?;
+    module_internal(&mut module)?;
+    module_instant(&mut module)?;
+
+    module.function_meta(sleep)?.docs(docstring! {
```
udoprog, line 44: You don't have to convert everything into using `docstring!`, since `function_meta` already picks it up if it's part of the function definition. It was only the associated constants (like `"SECOND"`) that could be improved.

Moving the docstrings here would prevent them from being visible to Rust users, so please move them back to the function definition.

ramon-bernardo: Okay, I assumed that `rune::function` would handle something like that, but I decided to switch everything to docstrings. I'll move it to a `Rust` instead.

udoprog (approved):
You're likely to have some clippy issues that needs to be resolved, but looks good. Thank you!

crates/rune-modules/src/time.rs
```diff
@@ -29,53 +29,599 @@
 //! }
 //! ```
 
-use rune::{Any, ContextError, Module};
+use rune::{
+    docstring,
+    runtime::{Mut, VmError, VmResult},
+    vm_panic, vm_try, Any, ContextError, Module,
+};
```
udoprog, line 36: Imports are broken up per-module in the rest of the project. It leads to cleaner diffs.

crates/rune-modules/src/time.rs
```diff
@@ -29,53 +29,599 @@
...
+    /// interval.tick().await;
+    /// println!("approximately 0ms have elapsed. The first tick completes immediately.");
+    /// interval.tick().await;
+    /// interval.tick().await;
+    ///
+    /// println!("approximately 20ms have elapsed...");
+    /// ```
+    pub async fn tick(mut internal: Mut<Interval>) {
+        internal.inner.tick().await;
+    }
```
ramon-bernardo, line 437: Doc for ticket fn will be applied? It doesn't seem right to me

ramon-bernardo:
Added methods to convert between the `Duration` type, `std::time::Duration` and `tokio::time::Duration`. I'm working on the http module, which necessitates handling conversions between different types.

crates/rune-modules/src/time.rs
```diff
@@ -29,53 +29,659 @@
...
+    ///
+    /// ```
+    /// use time::Duration;
+    ///
+    /// let duration = Duration::from_secs(5);
+    /// let std_duration = duration.into_std();
+    /// ```
+    pub fn into_std(self) -> std::time::Duration {
+        std::time::Duration::new(self.inner.as_secs(), self.inner.subsec_nanos())
+    }
```
ramon-bernardo, line 415: Considering this, we could make all methods private and eliminate the `keep` parameters (`rune::function`), handling the conversion between Duration's externally instead.

udoprog:
So some change is needed to allow a `Duration` to be stored as a constant value.

We currently don't seem to support it during compilation because internally we convert it to `ConstValue`. This is probably not necessary for types which are only used during compilation, but is needed if they are stored in the `Unit` since that needs to be `Sync`. An alternative might be to make it possible to store constant values of external types in some other way. Since they don't need to be reference-counted, they could be implemented in a thread-safe way.

crates/rune-modules/src/time.rs
udoprog: I'm fine either way. The `rune-modules` crate is currently unstable, so deciding what should and should not be public is up for future debate.

I have a minor preference to err on the side of caution just allowing for conversion seems like a good approach since you don't expect the rune-specific methods to be easily usable.

ramon-bernardo: Sure, that's fine with me as well!

udoprog:
I'd suggest you skip the constants for now until support for constructing them for custom types has been implemented so this and #819 can be merged as-is and further extended once support exists.

ramon-bernardo:
@udoprog ready for review! 😀

crates/rune-modules/src/time.rs
udoprog: The current project style would be:

```rust
use rune:{docstring, vm_panic, Any, ContextError, Module};
use rune::runtime::{Mut, VmResult};
```

udoprog (changes requested):
Same here, please remove commented out code and maintain in a separate branch instead if you want to keep it around for the future.

udoprog (approved):

udoprog:
I used this as basis for #861, since Duration uses `u64` and it's used to represent the full span of a Duration like with `Duration::MAX`. This made it really awkward and even impractical to interchangeably use the type for boundary values. Or, values which overflows `i64`.

I also added methods (like `as_secs` and more conversion methods) which were used in upstream test cases so we can adopt them more closely.

PR #819, opened 2024-09-02
modules: Expand http module with put, delete, head, auth and more

ramon-bernardo:
This PR significantly enhances the HTTP module by introducing...

### METHODS
- Put, Delete, Head, Form, Auth's, Timeout.

This PR requires the changes from [#815](https://github.com/rune-rs/rune/pull/815) to be merged first, as it includes necessary updates to the time module.

### TODO:
- Improve docs for the new methods with usage examples

### DIFF
--- a/crates/rune-modules/src/http.rs
+++ b/crates/rune-modules/src/http.rs
@@ -48,9 +48,13 @@
 //! }
 //! ```
 
+use core::hash::Hash;
+use core::cmp::Ordering;
+
 use rune::alloc::fmt::TryWrite;
-use rune::runtime::{Bytes, Formatter, Ref, VmResult};
-use rune::{docstring, Any, ContextError, Module, Value};
+use rune::runtime::{Bytes, Formatter, Hasher, Ref, VmResult};
+use rune::{docstring, item, Any, ContextError, Module, Value, ToConstValue};
+use rune::alloc::prelude::*;
 
 /// A simple HTTP module for Rune.
 ///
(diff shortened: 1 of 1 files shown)


crates/rune-modules/src/http.rs
```diff
@@ -284,6 +305,32 @@ impl RequestBuilder {
             request: self.request.body(bytes.into_std()),
         }
     }
+
+    #[rune::function]
+    fn form(self, params: HashMap<String, String>) -> Self {
```
ramon-bernardo, line 310: Rune HashMap; I don't see an issue with the conversion here.

crates/rune-modules/src/http.rs
```diff
@@ -241,6 +248,20 @@ impl RequestBuilder {
         }
     }
 
+    #[rune::function]
+    fn basic_auth(self, username: String, password: Option<String>) -> Self {
```
ramon-bernardo, line 252: `Option<String>` was used here because attempting to use `Option<&str>` results in a trait not impl error.

ramon-bernardo (commented):
Some points to conside, @udoprog

crates/rune-modules/src/http.rs
```diff
@@ -69,26 +69,1046 @@ pub fn module(_stdio: bool) -> Result<Module, ContextError> {
...
+            /// ```rune,no_run
+            /// use http::StatusCode;
+            ///
+            /// let status_code = StatusCode::NETWORK_AUTHENTICATION_REQUIRED;
+            /// ```
+        })?;
 
-    module.function_meta(Error::string_display)?;
     module.function_meta(StatusCode::string_display)?;
+    module.function_meta(StatusCode::as_integer)?;
```
ramon-bernardo, line 1001: By default, it would be `u16`, but since Rune uses `i64`, I opted for `Integer`.

crates/rune-modules/src/http.rs
```diff
@@ -177,6 +1197,37 @@ impl Response {
...
+    /// - The server didn't send a `content-length` header.
+    /// - The response is compressed and automatically decoded (thus changing
+    ///   the actual decoded length).
+    #[rune::function]
+    fn content_length(&self) -> Option<u64> {
+        self.response.content_length()
+    }
+
+    #[rune::function]
+    // TODO: create socket addr/ip addr into rune core
```
ramon-bernardo, line 1227: What if we had `socket/ip addr` in the core? Or perhaps a new net module?

crates/rune-modules/src/http.rs
```diff
@@ -192,6 +1243,69 @@ impl StatusCode {
         rune::vm_write!(f, "{}", self.inner);
         VmResult::Ok(())
     }
+
+    /// Returns the `Integer` corresponding to this `StatusCode`.
+    #[rune::function(instance)]
+    #[inline]
+    fn as_integer(&self) -> i64 {
+        self.inner.as_u16().into()
+    }
```
ramon-bernardo, line 1252: By default, it would be `u16`, but since Rune uses `i64`, I opted for `Integer`.

crates/rune-modules/src/http.rs
udoprog: Yeah. These should probably go into `std` which are the modules defined in the `rune` crate.

udoprog:
Similar issue here with constant values like in #815

crates/rune-modules/src/http.rs
udoprog: You can use `Option<Ref<str>>` to avoid cloning the string.

crates/rune-modules/src/http.rs
udoprog: :thinking:

To avoid copying it, you can use `rune::modules::collection::hash_map::HashMap`, but you'd have to make the methods you need to use `pub` (which I'm fine with) but then you'd have to ensure it implements `serde::Serialize`.

If you want to support anonymous objects though, this should be a reference to `rune::runtime::Object`, then calling this would support this syntax:

```rust
let request = /* .. */;

request = request.form(#{field: "value"});
```

I think that already implements `Serialize`.

crates/rune-modules/src/http.rs
```diff
@@ -69,26 +69,1046 @@ pub fn module(_stdio: bool) -> Result<Module, ContextError> {
...
+
+    module
+        .constant("CONTINUE", 100)
+        .build_associated::<StatusCode>()?
+        .docs(docstring! {
+            /// Status Code: Continue
+            ///
+            /// # Examples
+            ///
+            /// ```rune,no_run
```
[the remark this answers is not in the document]
ramon-bernardo: Only for constant?

crates/rune-modules/src/http.rs
ramon-bernardo: I'll put this on hold for now and focus on a new crate...  I'm considering `rune/modules/net`. After that, I'll implement `remote_addr`.

We could add support for both, `std::net` and `tokio::net`. What do you think?

crates/rune-modules/src/http.rs
ramon-bernardo: Alright, I've reverted to using `as_u16`.

crates/rune-modules/src/http.rs
```diff
@@ -69,26 +69,1046 @@ pub fn module(_stdio: bool) -> Result<Module, ContextError> {
...
-    module.function_meta(RequestBuilder::body_bytes)?;
+    module.function_meta(RequestBuilder::basic_auth)?;
+    module.function_meta(RequestBuilder::bearer_auth)?;
     module.function_meta(RequestBuilder::fetch_mode_no_cors)?;
+    module.function_meta(RequestBuilder::body_bytes)?;
+    module.function_meta(RequestBuilder::form)?;
+    module.function_meta(RequestBuilder::timeout)?;
+
+    module
+        .constant("CONTINUE", 100)
```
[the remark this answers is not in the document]
ramon-bernardo: Sure, I kept it as an integer due to `as_integer`... How could we implement `equality` here? For example, `StatusCodeA == StatusCodeB`

crates/rune-modules/src/http.rs
ramon-bernardo: Oh, I always forget about `Ref<T>`

crates/rune-modules/src/http.rs
```diff
@@ -172,11 +1504,40 @@ impl Response {
...
+    }
+
+    /// Get the remote address of the response.
+    #[rune::function(instance)]
+    fn remote_addr(&self) -> Option<SocketAddr> {
+        self.response
+            .remote_addr()
+            .map(|addr| SocketAddr::from_std(addr))
+    }
 }
```
ramon-bernardo, line 1541: This requires https://github.com/rune-rs/rune/pull/820 to be merged first.

crates/rune-modules/src/http.rs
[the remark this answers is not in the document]
udoprog: You implement the `PartialEq` trait and `PARTIAL_EQ` protocol, like this: https://github.com/rune-rs/rune/blob/bfb57ce123f1c71580e38916b5a889583d686ce1/crates/rune/src/modules/object.rs#L58

crates/rune-modules/src/http.rs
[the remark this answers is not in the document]
udoprog: `no_run` should only be used for code which cannot run. So probably not only for constants.

crates/rune-modules/src/http.rs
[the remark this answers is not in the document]
ramon-bernardo: Alright, I'll implement it tomorrow.

crates/rune-modules/src/http.rs
ramon-bernardo: I am implementing the net module in https://github.com/rune-rs/rune/pull/820, initially with IpAddr and SocketAddr.

crates/rune-modules/src/http.rs
ramon-bernardo: Future work ;)

crates/rune-modules/src/http.rs
ramon-bernardo: Future work

ramon-bernardo:
@udoprog ready for review! 😀

crates/rune-modules/src/http.rs
```diff
@@ -69,26 +69,1359 @@ pub fn module(_stdio: bool) -> Result<Module, ContextError> {
...
+            /// Status Code: Continue
+            ///
+            /// # Examples
+            ///
+            /// ```rune
+            /// use http::StatusCode;
+            ///
+            /// let status_code = StatusCode::CONTINUE;
+            /// ```
+        })?;
```
udoprog, line 119: The constants still won't work right since we don't have support for constructing constant values of external objects? So you'll have to remove them for now.

crates/rune-modules/src/http.rs
```diff
@@ -172,15 +1515,46 @@ impl Response {
...
     }
+
+    /// Get the version of the response.
+    #[rune::function(instance)]
+    fn version(&self) -> Version {
+        let inner = self.response.version();
+        Version { inner }
+    }
+
+    // TODO: response headers
```
udoprog, line 1531: Trailing TODO

crates/rune-modules/src/http.rs
```diff
@@ -172,15 +1515,46 @@ impl Response {
...
+    ///
+    /// - The server didn't send a `content-length` header.
+    /// - The response is compressed and automatically decoded (thus changing
+    ///   the actual decoded length).
+    #[rune::function(instance)]
+    fn content_length(&self) -> Option<u64> {
+        self.response.content_length()
+    }
+
+    /* TODO: Make rune net module
```
udoprog, line 1545: Please don't keep commented out code around, just remove it. If you want to keep the code around for the future, maintain it in a separate branch.

crates/rune-modules/src/http.rs
```diff
@@ -284,6 +1786,49 @@ impl RequestBuilder {
             request: self.request.body(bytes.into_std()),
         }
     }
+
+    /* TODO: Make with support anonymous objects: future work.
```
udoprog, line 1790: Same here

udoprog (changes requested):

udoprog (approved):

udoprog:
I added support for external constants in #860 and I was using this change as a reference. So I've re-added the constants now that it's been merged.

There were also some documentation and missing protocol implementations. In particular, when you implement a trait you also have to implement the associated protocols.

PR #820, opened 2024-09-02
Create net module

ramon-bernardo:
Add a 'net' module, initially with SocketAddr and IpAddr types and their respective methods. TCP and UDP will be implemented in the future.

### DIFF
--- a/crates/rune/src/modules/mod.rs
+++ b/crates/rune/src/modules/mod.rs
@@ -32,6 +32,7 @@ pub mod io;
 pub mod iter;
 pub mod macros;
 pub mod mem;
+pub mod net;
 pub mod num;
 pub mod object;
 pub mod ops;
--- a/crates/rune/src/modules/net.rs
+++ b/crates/rune/src/modules/net.rs
@@ -1,0 +1,163 @@
+//! The networking module.
+
+use std::net;
+
+use crate as rune;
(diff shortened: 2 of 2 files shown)


udoprog (commented):
Cheers. So we're missing rune doctests, and the module has to be installed to be available. See where the `module` function is used for other modules.

ramon-bernardo:
@udoprog how do I implement generating an IpAddr from a string, int, array of ints, etc., similar to the IpAddr inner?

udoprog:
> how do I implement generating an IpAddr from a string, int, array of ints, etc., similar to the std IpAddr?

You add the relevant methods to perform the conversion. Like the `parse` associated method, possible through a trait.

crates/rune-modules/src/net.rs
```diff
@@ -0,0 +1,185 @@
...
+        Self { inner: addr }
+    }
+}
+
+/// An IP address, either IPv4 or IPv6.
+#[derive(Debug, Any)]
+#[rune(item = ::std::net)]
+pub struct IpAddr {
+    inner: net::IpAddr,
+}
```
ramon-bernardo, line 130: We need a way to create an `IpAddr`. What is a preference: enum or constructors?

The std is `IpAddr::V4(...)` or `V6(...)`

PR #827, opened 2024-09-21
Remove copypaste from docs of `VecDeque` and `HashSet`

AngelicosPhosphoros:
Seems that docs were copied from `HashMap`, I made it more appropriate to the actual types.


udoprog:
Thank you!

PR #899, opened 2025-01-05
Add some functions for Bytes

traceflight:
Hi @udoprog, this pr updates some functions of Bytes.

1. Adds `push`, `remove`, `insert`, `index_get`, `index_set` functions. 
2. Registers `first` function, it seems that it was forgotten to be registered.


udoprog:
The error is just a line ending at the end of the script, you can format them the way Rune wants it like this:
```
cargo run --bin rune -- fmt --verbose --recursive scripts
```

udoprog:
Thank you!

PR #904, opened 2025-01-25
Add optional static typing to functions and structs

TheNachoBIT:
This tiny PR adds the option for the parser to detect "-> Type" at the end of the function if its detected.

### DIFF
--- a/crates/rune/src/ast/item_fn.rs
+++ b/crates/rune/src/ast/item_fn.rs
@@ -65,12 +65,9 @@ pub struct ItemFn {
     pub name: ast::Ident,
     /// The arguments of the function.
     pub args: ast::Parenthesized<ast::FnArg, T![,]>,
-    /// The arrow token.
-    #[rune(option)]
-    pub arrow: Option<ast::Arrow>,
     /// The function type.
     #[rune(option)]
-    pub fn_type: Option<ast::Ident>,
+    pub fn_type: Option<(ast::Arrow, ast::Type)>,
     /// The body of the function.
     pub body: ast::Block,
     /// Opaque identifier for fn item.
--- a/crates/rune/src/ast/item_fn.rs
+++ b/crates/rune/src/ast/item_fn.rs
@@ -67,7 +67,7 @@ pub struct ItemFn {
     pub args: ast::Parenthesized<ast::FnArg, T![,]>,
     /// The function type.
     #[rune(option)]
-    pub fn_type: Option<(ast::Arrow, ast::Type)>,
+    pub output: Option<(T![->], ast::Type)>,
     /// The body of the function.
     pub body: ast::Block,
     /// Opaque identifier for fn item.
--- a/crates/rune/src/ast/item_fn.rs
+++ b/crates/rune/src/ast/item_fn.rs
@@ -38,8 +38,7 @@ fn ast_parse() {
     assert!(item.const_token.is_some());
 
     let item_with_type = rt::<ast::ItemFn>("pub async fn hello(foo, bar) -> Type {}");
-    assert!(item_with_type.arrow.is_some());
-    assert!(item_with_type.fn_type.is_some());
+    assert!(item_with_type.output.is_some());
 }
 
 /// A function item.
--- a/crates/rune/src/ast/item_fn.rs
+++ b/crates/rune/src/ast/item_fn.rs
@@ -36,6 +36,10 @@ fn ast_parse() {
     assert_eq!(item.attributes.len(), 1);
     assert!(item.async_token.is_none());
     assert!(item.const_token.is_some());
+
+    let item_with_type = rt::<ast::ItemFn>("pub async fn hello(foo, bar) -> Type {}");
+    assert!(item_with_type.arrow.is_some());
+    assert!(item_with_type.fn_type.is_some());
 }
 
 /// A function item.
@@ -61,6 +65,12 @@ pub struct ItemFn {
     pub name: ast::Ident,
     /// The arguments of the function.
     pub args: ast::Parenthesized<ast::FnArg, T![,]>,
+    /// The arrow token.
+    #[rune(option)]
+    pub arrow: Option<ast::Arrow>,
+    /// The function type.
+    #[rune(option)]
+    pub fn_type: Option<ast::Ident>,
     /// The body of the function.
     pub body: ast::Block,
     /// Opaque identifier for fn item.
--- a/crates/rune/src/ast/item_struct.rs
+++ b/crates/rune/src/ast/item_struct.rs
@@ -77,10 +77,7 @@ pub struct Field {
     pub visibility: ast::Visibility,
     /// Name of the field.
     pub name: ast::Ident,
-    /// Colon token for the optional type.
-    #[rune(option)]
-    pub colon: Option<ast::Colon>,
     /// The type.
     #[rune(option)]
-    pub field_type: Option<ast::Ident>,
+    pub field_type: Option<(ast::Colon, ast::Type)>,
 }
--- a/crates/rune/src/ast/item_struct.rs
+++ b/crates/rune/src/ast/item_struct.rs
@@ -79,5 +79,5 @@ pub struct Field {
     pub name: ast::Ident,
     /// The type.
     #[rune(option)]
-    pub field_type: Option<(ast::Colon, ast::Type)>,
+    pub ty: Option<(T![:], ast::Type)>,
 }
--- a/crates/rune/src/ast/item_struct.rs
+++ b/crates/rune/src/ast/item_struct.rs
@@ -24,12 +24,13 @@ fn ast_parse() {
     rt::<ast::Field>("a");
     rt::<ast::Field>("#[x] a");
 
-    rt::<ast::ItemStruct>(r"
+    rt::<ast::ItemStruct>(
+        r"
         struct Foo { 
             a: i32, 
             b: f64, 
             c: CustomType, 
-        }"
+        }",
     );
 }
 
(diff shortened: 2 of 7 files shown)


TheNachoBIT:
Update: I've discovered that structs do not have static typing aswell (again, this is in the parser, not in the VM).

So i've made a little update that adds optional static typing to the struct's fields :D

crates/rune/src/ast/item_struct.rs
```diff
@@ -79,8 +79,8 @@ pub struct Field {
     pub name: ast::Ident,
     /// Colon token for the optional type.
     #[rune(option)]
-    pub colon: ast::Colon,
+    pub colon: Option<ast::Colon>,
```
udoprog, line 82: Type information should probably be in a single component, since not both the colon and type are optional.

So something like:

```
pub ty: Option<(Token![:], ast::Type)>
```

I'm not sure what `ast::Type` should be at the moment, but at least define it as something other than `ast::Ident`. One thing it definitely should support is `ast::Path` since the type might not be immediately defined.

udoprog (changes requested):
Cheers!

Could you also add the code which rejects the relevant items during indexing in case type information is specified? Otherwise the compiler will silently accept this even though it's subject to change which could accidentally allow breaking changes to sneak through.

That is currently checked in places like these: https://github.com/rune-rs/rune/blob/a833d25625fde93de2120ec3e6bca6e6fde293ad/crates/rune/src/indexing/index.rs#L231

crates/rune/src/ast/item_fn.rs
```diff
@@ -61,6 +65,12 @@ pub struct ItemFn {
     pub name: ast::Ident,
     /// The arguments of the function.
     pub args: ast::Parenthesized<ast::FnArg, T![,]>,
+    /// The arrow token.
+    #[rune(option)]
+    pub arrow: Option<ast::Arrow>,
```
udoprog, line 70: Same here.

crates/rune/src/ast/item_struct.rs
TheNachoBIT: Maybe `ast::Type` could've be an enum like this:

```rs
pub enum Type {
    Path(Path),
    Self(SelfType),
}
```

crates/rune/src/ast/item_fn.rs
TheNachoBIT: Sure! No problem!
I'll work on it :D

TheNachoBIT: Ig that should've be it c:

crates/rune/src/ast/item_struct.rs
TheNachoBIT: I've created "ast::Type" in the recent commit, i hope u like it :D

TheNachoBIT: (+ made the fixes that you suggested)

crates/rune/src/ast/rn_type.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// A type, used for static typing.
+#[derive(Debug, TryClone, PartialEq, Eq, ToTokens, Spanned)]
+#[non_exhaustive]
+pub enum Type {
+	/// If the type is an identifier or a path.
+	Path(ast::Path),
+	/// If the type is "Self".
+	SelfType(T![Self]),
```
udoprog, line 18: This should already be supported by `ast::Path` (see `PathSegment`).

crates/rune/src/ast/item_fn.rs
```diff
@@ -65,12 +65,9 @@ pub struct ItemFn {
     pub name: ast::Ident,
     /// The arguments of the function.
     pub args: ast::Parenthesized<ast::FnArg, T![,]>,
-    /// The arrow token.
-    #[rune(option)]
-    pub arrow: Option<ast::Arrow>,
     /// The function type.
     #[rune(option)]
-    pub fn_type: Option<ast::Ident>,
+    pub fn_type: Option<(ast::Arrow, ast::Type)>,
```
udoprog, line 70: I think `T![=>]` should work instead of `ast::Arrow`?

udoprog: Also this should probably be named `output`, see `Signature` in [`syn::Signature`](https://docs.rs/syn/latest/syn/struct.Signature.html). Adopting a `ReturnType` enum might even be an idea. The "type of the function" which `fn_type` hints as would be [something else](<https://rune-rs.github.io/play/?c=Zm4gZmliKCkgewp9CgpwdWIgZm4gbWFpbigpIHsKICAgIGxldCBhID0gZmliIGlzIHN0ZDo6b3BzOjpGdW5jdGlvbjsgLy8gdGhlIGZ1bmN0aW9uIHR5cGUKICAgIGxldCBiID0gZmliKCkgaXMgc3RkOjp0dXBsZTo6VHVwbGU7IC8vIHRoZSBvdXRwdXQKICAgIChhLCBiKQp9Cg%3D%3D>) on the language level.

crates/rune/src/ast/item_struct.rs
```diff
@@ -77,10 +77,7 @@ pub struct Field {
     pub visibility: ast::Visibility,
     /// Name of the field.
     pub name: ast::Ident,
-    /// Colon token for the optional type.
-    #[rune(option)]
-    pub colon: Option<ast::Colon>,
     /// The type.
     #[rune(option)]
-    pub field_type: Option<ast::Ident>,
+    pub field_type: Option<(ast::Colon, ast::Type)>,
```
udoprog, line 82: Same deal here, should prefer mimicking syn and calling it `ty`: <https://docs.rs/syn/latest/syn/struct.Field.html>. I would use `type` since everything else is more verbose, but since it is a reserved keyword `ty` is the convention used elsewhere. The `field_` prefix is superfluous since it's already part of the type called `Field`.

crates/rune/src/ast/rn_type.rs
TheNachoBIT: Oooh nice! I guess i forgot about that part, thank you for letting me know! I'll fix that rq :D

crates/rune/src/ast/item_fn.rs
TheNachoBIT: Sure! I'll do both of these + the "ty" in Struct!

TheNachoBIT: Done!

crates/rune/src/ast/rn_type.rs
```diff
@@ -14,17 +14,14 @@ fn ast_parse() {
...
 	/// If the type should return nothing (a.k.a the "never" type in Rust).
 	Bang(T![!]),
 }
 
 impl Parse for Type {
     fn parse(p: &mut Parser<'_>) -> Result<Self> {
         let segment = match p.nth(0)? {
-            K![Self] => Self::SelfType(p.parse()?),
-            K![ident] => Self::Path(p.parse()?),
+            K![ident] | K![Self] => Self::Path(p.parse()?),
```
udoprog, line 24: Should probably just use `ast::Path` parsing as fallback, since the other one is the special case.

udoprog (approved):

udoprog (commented):
I think the last required component is indexing rejection.

crates/rune/src/ast/rn_type.rs
TheNachoBIT: Oh, wait a minute, does path supports parenthesis?
I think i forgot to check what happens if you use a tuple as a return type.

TheNachoBIT: There, i've tested and i had to add tuple support.
I think you shouldn't fallback to just `ast::Path`, but let me know what you think c:

TheNachoBIT:
There, i fixed it and i think its now ready to run the tests :D

crates/rune/src/ast/rn_type.rs
udoprog: There's a lot of options missing, you should just use Path as fallback: https://github.com/rune-rs/rune/blob/a833d25625fde93de2120ec3e6bca6e6fde293ad/crates/rune/src/ast/path.rs#L211

udoprog:
If the type parameters are set, they still need to be rejected during indexing:
Here: https://github.com/rune-rs/rune/blob/a833d25625fde93de2120ec3e6bca6e6fde293ad/crates/rune/src/indexing/index.rs#L231
And here: https://github.com/rune-rs/rune/blob/a833d25625fde93de2120ec3e6bca6e6fde293ad/crates/rune/src/indexing/index.rs#L922

You can see other types of errors raised in there. Please also add the relevant test cases to make sure type annotations are rejected. You can see for example the attribute rejection test here to see how it can be done: https://github.com/rune-rs/rune/blob/a833d25625fde93de2120ec3e6bca6e6fde293ad/crates/rune/src/tests/attribute.rs#L92

crates/rune/src/ast/rn_type.rs
TheNachoBIT: Something like this is ok?
```rs
impl Parse for Type {
    fn parse(p: &mut Parser<'_>) -> Result<Self> {
        let segment = match p.nth(0)? {
            K![Self] => Self::Path(p.parse()?),
            K![!] => Self::Bang(p.parse()?),
            K!['('] => Self::Tuple(p.parse()?),
            _ => ast::Path::parse()?,
        };

        Ok(segment)
    }
}
```

TheNachoBIT:
Sure sure! I'll try that :D

crates/rune/src/ast/rn_type.rs
udoprog: Yes!

TheNachoBIT: Nice! I'll send a commit with this rq then! :D

TheNachoBIT:
There, i *think* i've added indexing rejection successfully.

udoprog (approved):
There are a few more places you could add if you wanted to, declarations and function arguments come to mind. But looks good for now!

udoprog:
Yeah, of course! Thank you!

PR #915, opened 2025-03-06
Fix build

dstoza:
Mostly fixing some new clippy diagnostics, but I also had to tweak a doc comment and update the `wasm-bindgen` version to fix an error inside the `#[wasm_bindgen]` macro.


crates/rune-alloc/src/ptr.rs
```diff
@@ -63,7 +63,7 @@ cfg_if! {
     if #[cfg(rune_nightly)] {
         #[inline(always)]
         pub(crate) unsafe fn sub_ptr<T>(from: *const T, to: *const T) -> usize {
-            from.sub_ptr(to)
+            from.offset_from_unsigned(to)
```
udoprog, line 66: I was trying to figure out what was going on since this method is not listed on Rust doc yet, when I realized ptr_sub_ptr is [just about to be stabilized](https://github.com/rust-lang/rust/issues/95892).

udoprog:
Thank you!

PR #950, opened 2025-06-17
Refine support for workspaces

dstoza:
This PR refines the existing workspace support to closely align with the typical [Rust package layout](https://doc.rust-lang.org/cargo/guide/project-layout.html).

For a full description of the workspace auto-detection based on package structure, see the new [section of the book](https://github.com/rune-rs/rune/pull/950/files#diff-7e596e72c5ddff206d81244ae5fa64f25d5c9dcd5eb10dc48224ad812ff5dd86).

On top of the new tests for workspace handling, the sample workspace created for the test was verified against the CLI commands for `check`, `doc`, `test`, `bench`, `run`, and `fmt` to make sure the correct workspace members were picked up.


udoprog:
Thanks!

I think you discussed the changes a bit on Discord, and they look good to me. Could you also go over what has changed in the description of the issue making it easier to keep a record?

crates/rune/src/workspace/manifest.rs
```diff
@@ -214,14 +233,37 @@ impl Package {
...
+        let mut output = Vec::new();
+
+        if let (Some(path), true) = (&self.root, self.auto_srcs) {
+            let path = path.join("src");
+            let files = find_rune_files(&path)?;
+            for file in files {
+                let (path, name) = file?;
+                if ["lib", "main"].iter().any(|entry| *entry == name) {
+                    output.try_push(Found {
+                        kind: FoundKind::Source,
```
udoprog, line 255: One thing we might want to be careful about is that a `lib` and a `main` are two different kinds of entrypoints. One is a binary named the same as the crate and if we follow Rust autosource conventions there can be more binaries under a `src/bin` folder. The other is a library.

I'm not sure this matters, but now might be a good time to set up workspace tests to make sure things work as intended even when many things are named the same.

dstoza: Yeah, I agree. I'll try to formalize this to match Rust conventions and change the existing workspace stuff in the repo to match next week. I just wanted to make sure I was on the right track first 🙂

udoprog (approved):
Looks great. Thank you for putting this together! I'll merge this when I get back feel free in case you want to tinker some more.

dstoza:
The `--package` flag was the last thing I wanted to do (it was bugging me that it didn't exist, even if I don't think it will be used that often), but I figure I can rebase this after my clippy PR goes in so this PR is a bit cleaner in the log.

udoprog:
Thank you!

PR #954, opened 2025-06-27
Fix clippy lints

dstoza:
The latest set of clippy diagnostics includes `uninlined_format_args`, which hits a lot of places in the codebase 😅

I also added `clippy::unnecessary_operation` to the `#[allow(clippy::no_effect)]` in `crates/rune-alloc/src/hashbrown/map.rs` and added an `#[expect(clippy::mut_from_ref)]` in `crates/rune/src/hir/arena.rs` (which seems to be the intended behavior).


udoprog:
LGTM. I'm traveling at the moment but I'll be back in a week!

PR #957, opened 2025-08-13
Add constant vec example

SWW13:
I struggled quite a while trying create a vec constant with v0.14.0 as neither `vec![1, 2, 3]`, `[1, 2, 3]`, `&[1, 2, 3]`, `ConstValue::from(...)` and many more didn't work.

### DIFF
--- a/crates/rune/src/module/module.rs
+++ b/crates/rune/src/module/module.rs
@@ -319,6 +319,9 @@ impl Module {
     ///         /// A global ten value.
     ///     });
     ///
+    /// module.constant("VEC", [1, 2, 3].as_slice())
+    ///     .build()?;
+    ///
     /// module.constant("TEN", 10)
     ///     .build_associated::<MyType>()?
     ///     .docs(docstring! {


udoprog:
Thanks!

SWW13:
Hmm, looks like this only works because it creates a `Bytes` constant. Any idea how to create a const integer vec?

```
error: literal out of range for `u8`
  --> test.rs:23:32
   |
23 |     module.constant(["TEST"], [501, 502].as_slice()).build()?; // nop
   |                                ^^^
   |
   = note: the literal `501` does not fit into the type `u8` whose range is `0..=255`
   = note: `#[deny(overflowing_literals)]` on by default
```

udoprog:
The most obvious approach would be to construct a `ConstValue` somehow. I don't think there's a constructor available right now but you could add it.

Another could be to implement `ToConstValue` for `Vec<T> where T: ToConstValue`?

SWW13:
Even with a `ConstValue` value this does not seem to work (besides the issue of creating one from e.g. a `Vec<u64>`):

```Rust
    module
        .constant("TEST", ConstValue::bytes(&[1, 2]).unwrap())
        .build()?;
```

```
error[E0277]: the trait bound `ConstValue: rune::runtime::TypeOf` is not satisfied
   --> test.rs:25:27
    |
25  |         .constant("TEST", ConstValue::bytes(&[1, 2]).unwrap())
    |          --------         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `rune::runtime::TypeOf` is not implemented for `ConstValue`
    |          |
    |          required by a bound introduced by this call
    |
    = help: the following other types implement trait `rune::runtime::TypeOf`:
              &T
              &mut T
              ()
              (A, B)
              (A, B, C)
              (A, B, C, D)
              (A, B, C, D, E)
              (A, B, C, D, E, F)
            and 88 others
note: required by a bound in `rune::Module::constant`
   --> .cargo/git/checkouts/rune-ad90a1eda356f8c2/61d267a/crates/rune/src/module/module.rs:334:23
    |
332 |     pub fn constant<N, V>(&mut self, name: N, value: V) -> ModuleConstantBuilder<'_, N, V>
    |            -------- required by a bound in this associated function
333 |     where
334 |         V: TypeHash + TypeOf + ToConstValue,
    |                       ^^^^^^ required by this bound in `Module::constant`

error[E0599]: the method `build` exists for struct `ModuleConstantBuilder<'_, &str, ConstValue>`, but its trait bounds were not satisfied
   --> test.rs:26:10
    |
24  | /     module
25  | |         .constant("TEST", ConstValue::bytes(&[1, 2]).unwrap())
26  | |         .build()?;
    | |         -^^^^^ method cannot be called on `ModuleConstantBuilder<'_, &str, ConstValue>` due to unsatisfied trait bounds
    | |_________|
    |
    |
   ::: .cargo/git/checkouts/rune-ad90a1eda356f8c2/61d267a/crates/rune/src/runtime/const_value.rs:251:1
    |
251 |   pub struct ConstValue {
    |   --------------------- doesn't satisfy `ConstValue: TypeHash` or `ConstValue: rune::runtime::TypeOf`
    |
    = note: the following trait bounds were not satisfied:
            `ConstValue: TypeHash`
            `ConstValue: rune::runtime::TypeOf`
```

udoprog:
The `TypeOf` and `TypeHash` bounds in this instance seem unnecessarily strict, since the type information *is* available at runtime. This might be something you could relax. Alternatively you need some type which you can wrap the vector data in which has the correct type information.

So this is not immediately supported right now. Some work has to be put into it. If you're interested in doing so, please open up a specific issue for it and we can continue working there.

PR #980, opened 2025-11-25
Gradual Typing - First Implementation

Mootikins:
First pass implementation on gradual typing. I have a personal project (currently private) that would benefit greatly from gradual typing, so I thought I'd try to reignite the conversation with an initial attempt. Open to feedback and criticism to push this forwards.

## Key Features

- Type annotations for function parameters and return types
- Type inference with Hindley-Milner style algorithm
- Gradual semantics - untyped code treated as Any type
- Optional strict mode for type errors vs warnings
- Type extraction API for embedders

## Files Modified

- Core: compile/typeck.rs, compile/type_info.rs
- AST: ast/fn_arg.rs - added typed parameter support
- Runtime: runtime/unit.rs - type extraction methods
- Indexing: index/index.rs, index/mod.rs - type capture
- Tests: tests/gradual_typing.rs, tests/type_extraction.rs

## Compatibility

- 100% backwards compatible - existing code unchanged
- Feature-gated behind gradual-typing flag
- Zero runtime overhead - compile-time type checking only
- Gradual adoption - mix typed and untyped code

## Testing

- 29 new tests covering type checking, inference, and API
- All existing tests pass (339/339)
- Clippy clean - all linting issues resolved

## Usage Example

```rune
fn add(a: i64, b: i64) -> i64 {
    a + b  // Type checked
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,8 +6,11 @@ members = [
     "examples",
     "benches",
     "tools/*",
-    "no-std",
 ]
+
+# no-std example requires nightly and defines its own #[panic_handler],
+# which conflicts with std. Run separately: cargo run --manifest-path=no-std/Cargo.toml --example minimal
+exclude = ["no-std"]
 
 default-members = [
     "crates/*",
--- a/crates/rune/src/runtime/unit.rs
+++ b/crates/rune/src/runtime/unit.rs
@@ -449,7 +449,7 @@ impl<S> Unit<S> {
             let type_info = field_types_map
                 .get(field_name.as_ref())
                 .and_then(|ty_opt| *ty_opt)
-                .map(|s| parse_type_string(s));
+                .map(parse_type_string);
 
             fields
                 .try_push(FieldInfo {
--- a/no-std/examples/minimal.rs
+++ b/no-std/examples/minimal.rs
@@ -67,10 +67,15 @@ fn inner_main() -> rune::support::Result<i32> {
     let context = rune::Context::with_default_modules()?;
     let runtime = Arc::try_new(context.runtime()?)?;
 
+    // Test gradual typing in no-std environment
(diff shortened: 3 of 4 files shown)


udoprog:
Very cool!

The overall implementation looks good to me. But as a starter, please remove the feature flag in favor of just checking a compile option since it can cause discrepancies in runtime behavior during feature unification. As-in, one project activating the feature can transitively change the behavior of another.

I'll sweep over the nits next.

udoprog (commented):
One thing I'm curious about is if there's anything that prevents you from doing type-checking during [ast lowering](https://github.com/rune-rs/rune/blob/f67a45eac99ed609facc2b11d75352e556ffa0a0/crates/rune/src/hir/lowering.rs) which already walks the full ast?

This is an instance where we [already query for and use type hashes](https://github.com/rune-rs/rune/blob/f67a45eac99ed609facc2b11d75352e556ffa0a0/crates/rune/src/hir/lowering.rs#L1807) which in essence is a form of type checking. With type annotations I think you could resolve those and apply the `typeck` logic you've build there.

udoprog:
It would be nice if you could also break out the parsed type annotations and associated tests into a separate PR that just rejects type annotations as an error during indexing into a separate PR to make the implementation phase smaller to iterate on.

Mootikins:
@udoprog I think I covered your feedback. I also updated the ACE highlighting regex for the type hints.

I made a few scripts for running the CI checks locally if you'd like me to make a separate PR for those (or include them here).

udoprog:
So the point of integrating type checking into hir lowering is because all the work of doing type meta lookups is already done during lowering, and that it can be extended with the updated AST to include more items.

So the latest change just moved the call into typeck into item_fn which duplicates the work of walking over the entire AST. I would like to see this integrated into the lowering.

PR #983, opened 2026-01-03
Implement bitwise operations

Ryan1729:
I ran into the "op not supported yet" error while trying to use numbers as bitflags in Rune. I noticed that adding support for those operations seemed to be pretty straightforward, hence this PR.


Ryan1729:
After fixing the formatting, I managed to get the CI running on my fork by enabling Actions then making [a PR there](https://github.com/Ryan1729/rune/pull/1), and the checks passed. So I'm now confident they will pass here.

PR #1014, opened 2026-03-26
Change mod's Ident to Path for more advanced module resolution.

djgaven588:
The goal of this change is to allow more advanced path traversal for file resolvers, specifically allowing the following:
``mod ::VoxelEras::DestinyHailstorm::settings;``
Where ``::VoxelEras::DestinyHailstorm`` refers to an absolute path of a given mod, or alternatively:
``mod crate::settings``
Which refers to an absolute path within the context of the given mod.

This still allows modules to be resolved "locally" like so:
``mod settings;``
Where the file path can be used to determine the module's location.

In Rhai (with my custom file resolver), this is:
``import "/VoxelEras:DestinyHailstorm/settings" as settings;``
Or locally:
``import "./settings" as settings;``

I'm not sure if this would have any side effects, but it does seem to compile and work fine in my project so far.

### DIFF
--- a/crates/rune/src/ast/item_mod.rs
+++ b/crates/rune/src/ast/item_mod.rs
@@ -27,7 +27,7 @@ pub struct ItemMod {
     /// The `mod` keyword.
     pub mod_token: T![mod],
     /// The name of the mod.
-    pub name: ast::Ident,
+    pub name: ast::Path,
     /// The optional body of the module declaration.
     pub body: ItemModBody,
     /// The id of the module item.
--- a/crates/rune/src/indexing/indexer.rs
+++ b/crates/rune/src/indexing/indexer.rs
@@ -446,7 +446,13 @@ impl Indexer<'_, '_> {
         ast: &mut ast::ItemMod,
         docs: &[Doc],
     ) -> compile::Result<()> {
-        let name = ast.name.resolve(resolve_context!(self.q))?;
+        let original_name = ast
+            .name
+            .resolve(resolve_context!(self.q))?
+            .try_to_string()
+            .unwrap();
+        let name = Box::try_new(original_name.split("::").last().unwrap())?;
+
         let visibility = ast_to_visibility(&ast.visibility)?;
         let guard = self.items.push_name(name.as_ref())?;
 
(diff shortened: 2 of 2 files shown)


djgaven588:
Hmm, while it seemed to work on the surface, it only seems to have worked with the prior usage, not the new use case I tried to add.

udoprog:
Thanks for the suggestion.

Rust is an important touchstone when it comes to design, and since this very differently with regards to modules I'm probably not going to want to include it.

djgaven588:
> Thanks for the suggestion.
> 
> Rust is an important touchstone when it comes to design, and since this very differently with regards to modules I'm probably not going to want to include it.

Would you have any suggestion as to how to achieve this within Rust's design ideas?

In Rust you're able to use the Cargo.toml to bring in external packages, but there is no equivalent here, as well as the ``mod`` keyword working entirely different than ``use``, as you're unable to specify submodules like ``my_module::my_submodule`` when you're bringing them into another script, since you can only use a single identifier ``my_submodule``.

udoprog:
Since I don't know exactly what you're doing, it looks like you want to add support for the `#[path = ..]` attribute to specify a path.

```
#[path = "my/custom/path.rn"]
mod imported;
```

Then passing that into the module resolver that determines which source is loaded from where.

Or maybe adding support for crates which are essentially code modules compiled from different places and imported during compile time into some namespace resolved by `imported` or `::imported`.

If you're using module declarations to "import code" you end up with a ton of duplicated code instead of the modules being imported as you expect them to.

djgaven588:
I think the path attribute would be kinda what I'm looking to do, as I'm wanting to be able to do the following:
```
- [Folder] Voxel Eras (Identifier "VoxelEras:DestinyHailstorm")
    - mod.rn (#[path = "./scripts/utils.rn"] mod utils;)
    - [Folder] scripts
        - utils.rn
- [Folder] Test Mod
    - mod.rn (#[path = "/VoxelEras:DestinyHailstorm/scripts/utils.rn] mod other_utils;)
```
The idea here being that each of the top level folders are a (game) mod, with local paths referring to within the same (game) mod, and absolute paths allow accessing another (game) mod's scripts. (In reality these folders can be in various locations and various formats, like zips and Steam Workshop mods, so this folder structure is somewhat abstracted away)

Code duplication isn't much of a concern, these are typically small scripts with 1-2 modules brought in.

PR #1017, opened 2026-05-19
fix(compile): propagate divergence from expr_match when all arms diverge

### LINKED ISSUE #1016: Compiler error `No address for need defer(0:0 (return value))` when all match arms diverge
## Summary

Using `return` inside all arms of a `match` expression in an async function produces a compiler error:

```
error: No address for need defer(0:0 (return value))
```

The same pattern works correctly with `if/else` (which already has divergence tracking), but `expr_match` always reports convergence even when all arms diverge.

## Minimal reproduction

```rune
pub async fn run(db, i) {
    let result = some_fallible_call(db).await;

    match result {
        Err(e) => {
            return Err(e)
        },
        _ => {
            return some_other_call(db, i).await
        }
    }
}

async fn some_fallible_call(db) {
    Ok(())
}

async fn some_other_call(db, i) {
    Ok(())
}
```

Compiling this produces:
```
error: No address for need defer(0:0 (return value))
```

## Expected behavior

The code should compile successfully. Every arm produces a value via `return`, so the match expression as a whole diverges and the compiler should not attempt to allocate a return-value address for it.

## Workaround

Remove explicit `return` from match arms and use the match as an implicit return expression:

```rune
pub async fn run(db, i) {
    let result = some_fallible_call(db).await;

    match result {
        Err(e) => {
            Err(e)
        },
        _ => {
            some_other_call(db, i).await
        }
    }
}
```

## Root cause

In `crates/rune/src/compile/v1/assemble.rs`, `expr_match` always returns `Ok(Asm::new(span, ()))` (converging), regardless of whether all branch bodies diverge. When the match is used as the expression inside a `return_()` call, the `Defer` needs is never allocated (because no branch writes a value — they all diverge via `return`), and the subsequent `needs.addr()` call fails.

`expr_if` already handles this correctly by tracking `all_diverging` and returning `Asm::diverge(span)` when appropriate. The same logic is missing from `expr_match`.

## Environment

- rune version: 0.14.x (confirmed on 0.14.0 and 0.14.1)
- Also affects non-async functions with the same pattern

vponomaryov:
When all match arms diverge (every arm contains a `return` statement), the `expr_match` now returns `Asm::diverge` instead of always reporting convergence.

Previously, the parent `return_` function would attempt to read the address of a Defer needs that was never allocated, producing the following error:

    "No address for need defer(0:0 (return value))"

This mirrors the existing divergence tracking already present in `expr_if`.

Fixes: #1016

### DIFF
--- a/crates/rune/src/compile/v1/assemble.rs
+++ b/crates/rune/src/compile/v1/assemble.rs
@@ -2646,6 +2646,7 @@ fn expr_match<'a, 'hir>(
         cx.asm.jump(&end_label, span)?;
     }
 
+    let mut all_diverge = is_irrefutable;
     let mut it = hir.branches.iter().zip(branches).peekable();
 
     while let Some((branch, (label, scope))) = it.next() {
@@ -2654,8 +2655,11 @@ fn expr_match<'a, 'hir>(
         cx.asm.label(&label)?;
         let scope = cx.scopes.restore(scope);
 
-        if expr(cx, &branch.body, needs)?.converging() && it.peek().is_some() {
-            cx.asm.jump(&end_label, span)?;
+        if expr(cx, &branch.body, needs)?.converging() {
+            all_diverge = false;
+            if it.peek().is_some() {
+                cx.asm.jump(&end_label, span)?;
+            }
         }
 
         cx.scopes.pop(span, scope)?;
@@ -2665,7 +2669,12 @@ fn expr_match<'a, 'hir>(
 
     value.free()?;
     linear.free()?;
-    Ok(Asm::new(span, ()))
+
+    if all_diverge {
+        Ok(Asm::diverge(span))
+    } else {
+        Ok(Asm::new(span, ()))
+    }
 }
 
 /// Compile a literal object.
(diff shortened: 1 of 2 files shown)


udoprog:
Very cool! Could you add a test for this as well? A rune test file would be fine since this catches a miscompilation. Just the fact that it runs then means it should be fixed.

vponomaryov:
> Very cool! Could you add a test for this as well? A rune test file would be fine since this catches a miscompilation. Just the fact that it runs then means it should be fixed.

Added tests in a separate commit to be able to apply it to the compiler with and without the fix.

vponomaryov:
@udoprog 
What is the procedure for backporting it?
It must be backported to the `0.14.x`.
Also, the `0.14.x` must have an additional fix for the `if-else` which is ok in the `main`, but not ok in there.

udoprog:
Thank you!

Usually I just backport the fixes to the branch associated with a given release.

vponomaryov:
> Thank you!
> 
> Usually I just backport the fixes to the branch associated with a given release.

I can create a separate PR with the additional changes needed for the `0.14.x` branch.

udoprog:
Please do!

PR #1025, opened 2026-07-18
Pass values through to nested compile options

polina4096:
The value of a nested option was dropped during parsing, since `fmt.<option>=<value>` was split on `=` first. Only the option key was forwarded to the fmt option parser.

Example: `-O fmt.force-newline=false` actually set `force_newline = true`.


udoprog:
I'm fixing lints and updating dependencies for the project in #1027 which should resolve the build issues. Once it has landed, I'll rebase!

PR #1026, opened 2026-07-18
Implement support for passing custom indentation to the formatter

polina4096:
Currently, the formatter always indents with hard-coded 4 spaces. This PR adds a `fmt.indent` option (`-O fmt.indent=<number>` or `-O fmt.indent=tab`, default `4`) and makes the language server use the `tabSize` / `insertSpaces` values clients send with `textDocument/formatting` and `textDocument/rangeFormatting`, which were previously ignored.

Needs #1025 to be merged first, without it option values never reach the fmt parser.


udoprog:
Thanks, this is a nice change!

PR #1032, opened 2026-07-27
Fix statements such as `if cond { } (..)` being treated as call expressions

### LINKED ISSUE #1021: Statements like `if cond { } (..)` are incorrectly treated as function calls
But if we are writing in rust,

```rune
if cond { }
(x, y, z);
```
This should not be a function call.

And
```rune
(if cond { })
(x, y, z);
```
This should be a function call.

polina4096:
Statements beginning with a block-like expression (if, match, select, loop, or a plain block) now end at their closing brace, matching Rust: a following `(..)` or `[..]` starts a new statement instead of becoming a call or index on the block. Expression position is unchanged.

`let x = if c { f } else { g }();` still calls -- as are `.field`, `.await`, `?` and binary operators.

Implemented in both parsers. The AST parser now parses statements with `NOT_CALLABLE`, and the grammar parser gets an equivalent Callable flag.

Fixes #1021

### DIFF
--- a/crates/rune/src/ast/expr.rs
+++ b/crates/rune/src/ast/expr.rs
@@ -105,7 +105,9 @@ pub(crate) struct Callable(bool);
 pub(crate) const CALLABLE: Callable = Callable(true);
 
 /// Indicates that an expression should be treated as if it's *not* callable.
-/// This is used to solve otherwise parsing ambiguities.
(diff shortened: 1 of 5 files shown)


udoprog:
Very cool and a nice targeted fix. Thank you!

Issue #1040, opened 2026-08-03
Deeply nested expressions abort the process with a stack overflow at compile time


DRMacIver:
Compiling a deeply nested expression overflows the stack and aborts the process instead of returning a compile error:

```rust
use rune::{Context, Diagnostics, Source, Sources};

fn main() {
    let depth = 100_000;
    let src = format!("pub fn main() {{ {}0{} }}", "-(".repeat(depth), ")".repeat(depth));

    let context = Context::with_default_modules().unwrap();
    let mut sources = Sources::new();
    sources.insert(Source::memory(&src).unwrap()).unwrap();
    let mut diagnostics = Diagnostics::new();
    let _ = rune::prepare(&mut sources)
        .with_context(&context)
        .with_diagnostics(&mut diagnostics)
        .build();
}
```

Output:

```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The process is killed with SIGABRT, which a host embedding rune cannot catch, so script or REPL input with enough nesting takes the host process down. I'd expect a recursion-depth/nesting-limit error instead. Parenthesized groups `(((...0...)))` and other nested shapes abort the same way; the depth required depends on the available stack.

Tested on rune 0.14.2 and on current `main`.

This was another bug found with Hegel while preparing the follow-on PR discussed in #1030.

udoprog:
Cheers! This is actually being worked on in the v2 compiler (limiting stack use). I'll be afk until next week, but I'll get back to this!

Issue #1039, opened 2026-08-03
Compiling a `const` with overflowing integer arithmetic panics the compiler


DRMacIver:
Compiling a `const` whose initializer overflows `i64` panics the compiler on a debug build, and accepts the wrapped value on a release build:

```rust
use rune::{Context, Diagnostics, Source, Sources};

fn main() {
    let context = Context::with_default_modules().unwrap();
    let mut sources = Sources::new();
    sources
        .insert(Source::memory("pub const VALUE = 9223372036854775807 + 1;").unwrap())
        .unwrap();
    let mut diagnostics = Diagnostics::new();
    let _ = rune::prepare(&mut sources)
        .with_context(&context)
        .with_diagnostics(&mut diagnostics)
        .build();
}
```

On a debug build this panics:

```
thread 'main' panicked at src/compile/ir/eval.rs:84:57:
attempt to add with overflow
```

On a release build `build()` returns `Ok` and the constant takes the wrapped value.

The same overflowing arithmetic evaluated at runtime (e.g. `9223372036854775807 + 1` in a function body) returns an error rather than panicking, so I'd expect a `const` initializer to report an error here too instead of panicking (debug) or wrapping (release).

Tested on rune 0.14.2 and on current `main`.

This was another bug found with Hegel while preparing the follow-on PR discussed in #1030.

Issue #1030, opened 2026-07-26
Negating `i64::MIN` in a Rune script panics the host in debug and wraps in release


DRMacIver:
```rust
use rune::{Context, Diagnostics, Source, Sources, Vm};
use rune::termcolor::{ColorChoice, StandardStream};
use std::sync::Arc;

fn main() -> rune::support::Result<()> {
    let context = Context::with_default_modules()?;
    let runtime = Arc::new(context.runtime()?);

    let mut sources = Sources::new();
    sources.insert(Source::memory(
        r#"
        pub fn main() {
            let a = -9223372036854775808;
            -a
        }
        "#,
    )?)?;

    let mut diagnostics = Diagnostics::new();

    let result = rune::prepare(&mut sources)
        .with_context(&context)
        .with_diagnostics(&mut diagnostics)
        .build();

    if !diagnostics.is_empty() {
        let mut writer = StandardStream::stderr(ColorChoice::Always);
        diagnostics.emit(&mut writer, &sources)?;
    }

    let unit = result?;
    let unit = Arc::new(unit);
    let mut vm = Vm::new(runtime, unit);

    let output = vm.call(["main"], ())?;
    println!("{:?}", output);
    Ok(())
}
```

On a debug build this panics:

```
thread 'main' panicked at core/src/ops/arith.rs:729:1:
attempt to negate with overflow
```

On a release build it prints the unchanged value instead:

```
-9223372036854775808
```

For comparison, an integer overflow through `+` on the same VM (e.g. `9223372036854775807 + 1`) does not panic in either profile; `vm.call` returns `Err(VmError { ... kind: Overflow ... })`. Unary negation of `i64::MIN` is the odd one out: it panics the embedding host application in debug and produces a wrong value in release.

Tested on `rune` 0.14.2.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

udoprog:
> Happy to contribute the tests if you're interested.

Sounds interesting. Are there any downsides to incorporating hegel into our testing strategy?

DRMacIver:
> Are there any downsides to incorporating hegel into our testing strategy?

I'm obviously biased, but personally I don't think there are any major ones!

The main things:

1. Hegel tests are a bit slower than other tests because they run (by default) 100 test cases (and have a bit of overhead for test case generation). Still well under a second for the tests I've added here.
2. It adds another testing dependency
[...]

PR #1041, opened 2026-08-03
Add hegel property tests as a soft CI dependency

DRMacIver:
Follow up to #1030 which adds the Hegel tests that found that bug.

This includes a bunch of tests that I think are good tests but do not currently seem to find bugs in rune. Totally happy to ditch those if you prefer.

I've added this as a separate non-blocking job in CI, via a separate hegel_tests crate. I'm not particularly attached to this approach (it was Claude's suggestion for your requirements) so let me know if you want me to back some or all of the CI integration out.

NB this currently has workarounds to avoid hitting #1039 and #1040


udoprog:
All right, cheers!

